PR #38, opened 2026-09-06
feat(review): give the session one shape and an honest track

krambertech:
## ✨ What

Review is one screen again: the count and the exit share a line, the session track sits under them, and the whole card reveals the answer instead of a caption asking you to tap it. The grade strip is the same height before and after the reveal, so the card no longer shrinks and shoves the answer upward at the moment you are reading it.

A session is one batch of at most fifty. When more is due the end screen offers **Keep going** rather than announcing a number you have no way to act on.

## 🤔 Why

Five things were costing something on every card. The header spent two rows and about 90 px before the word. The grade bar mounted only on reveal, so the layout jumped exactly when attention landed on the answer. Reveal — the most-pressed control on the screen — was muted text at the foot of the card. The four scheduling dates FSRS computes for the grade buttons arrived in the queue payload and were thrown away. And on desktop the 240 px rail stayed up through the session while the phone hid its pill, which made focus a phone-only idea.

The deck name leaves the session for the end screen. It is chosen two taps earlier, it cannot change for the next eleven cards, and a long one squeezes the track down to nothing. The lantern stays in the header, because the flare on a good answer is the one warm thing on this screen.

The track is 8 px on the card's own width rather than a hairline on the screen edge: rounded corners and the status bar make an edge hairline read as a rendering fault rather than a thing. Its fill is ink rather than amber, which also puts `Progress` back inside the rule that amber is the flame, the one primary action and a due count.

## ✅ How to verify

- `pnpm verify` — formatting, lint, production build, typecheck, unit tests. Passing.
- `pnpm test:e2e` — 16 passed across Chromium, WebKit and mobile, including the capture-and-review flow that exercises reveal, grading and the end state.
- By hand: open `/review` on a desktop window and confirm the rail is gone; hover a grade to see what it schedules; reveal a card and confirm nothing moves except the answer arriving under the rule.

<details>
<summary><code>pnpm test:e2e</code></summary>

```text
Running 16 tests using 1 worker
  16 passed (2.1m)
```

</details>

## 📝 Notes

Two things the mockups showed that are deliberately not here, because each needs a feature rather than a layout: undo of a grade, which needs a server-side ungrade path, and "fix this card", which needs a card edit sheet that does not exist yet.

Prioritisation is the open question this change makes visible. The queue is still ordered by `due` ascending, and with the batch boundary now explicit it matters which fifty you get. That is its own ADR and its own pull request.


apps/web/src/client/routes/review.tsx
```diff
@@ -47,6 +47,17 @@ function Review() {
   const sessionTotal = items.length;
   const finished = queue.isSuccess && !current;
   const deckName = deck ? decks.data?.find((d) => d.id === deck)?.name : undefined;
+  // A session is one batch. What did not fit is offered as the next one rather than appended,
+  // because fifty cards at a sitting is already more than an evening wants.
+  const moreDue = Math.max(total - items.length, 0);
+
+  /** Start the next batch: refetch what is due now and begin its count from zero. */
+  const nextBatch = useCallback(() => {
```
krambertech: [Major] `setIndex(0)` runs before the invalidation/refetch completes, while React Query keeps the completed batch in `queue.data`. In a 51-card reproduction, I finished the first 50, held the refetch, and pressed **Keep going**; the UI immediately reopened the first reviewed card as `0 of 50`, allowing it to be revealed and graded again. Please fetch or clear/replace the old queue before resetting the index, keep a loading state visible in the meantime, and add a 51-card regression test for this transition.

apps/web/src/client/views/ReviewView.tsx
```diff
@@ -286,6 +300,18 @@ export function GradeBar({
               <span className={clsx("transition-opacity duration-150", saving && "opacity-0")}>
                 {g.label}
               </span>
+              {next && (
+                <span
+                  aria-hidden="true"
+                  className={clsx(
+                    "hidden h-4 text-2xs tabular-nums leading-4 text-faint opacity-0 transition-opacity duration-150",
```
krambertech: [Major] These intervals are functional text, but `text-faint` renders at 2.29:1 on the light hovered button and 2.24:1 in dark mode at 10 px, below Lymi's 4.5:1 floor. The same content is also `aria-hidden`, so screen-reader users never receive the scheduling information exposed on hover/focus. Please use a text token that passes in both themes and include the interval in the button's accessible name or description while keeping the visible treatment compact.

apps/web/src/client/views/ReviewView.tsx
```diff
@@ -231,24 +226,43 @@ export function ReviewCard({
...
   pendingRating,
   error,
   onGrade,
   className,
 }: GradeBarProps) {
+  const now = new Date();
+  if (!revealed) {
+    return (
+      <div className={clsx("grid h-[72px] shrink-0 place-items-center", className)}>
+        <p className="text-sm font-medium text-muted">Tap the card when you have it</p>
```
krambertech: [Minor] This replaces the previous desktop cue, `Tap card to reveal  Space`, with touch-only wording at every size. The Space shortcut still works, but it is no longer discoverable on the desktop flow that explicitly supports keyboard review. Please restore the `<Kbd>Space</Kbd>` hint on keyboard-sized layouts, or use input-neutral visible copy with the shortcut beside it.

krambertech (commented):
I exercised the review flow on live 320 px phone and 1440 px desktop layouts, in light and dark themes, with reduced motion, keyboard reveal/grading, and a 51-card queue. The focused layout and stable grade strip are a clear improvement. One batching correctness bug and one accessibility regression still need addressing; I also left a small desktop shortcut-discoverability note.

apps/web/src/client/routes/review.tsx
krambertech: Fixed in 07ef65d. You were right about the mechanism: React Query keeps serving the finished batch until the next fetch replaces it, so the index reset landed while `queue.data` still held the old 50.

`nextBatch` now awaits `refetchQueries` before touching the index, and the button holds a local `startingNext` loading state across the wait rather than keying off `queue.isFetching`.

Regression test added as `e2e/review-batching.spec.ts`: it builds a 51-card queue, grades all 50 through the UI, presses Keep going, and asserts the next card is not one of the 50 already graded. I checked it actually bites — against the old implementation it fails with "the next batch must not reopen a graded card". Runs Chromium-only and takes about 27s, since what it covers is the batch transition rather than how it renders.

apps/web/src/client/views/ReviewView.tsx
krambertech: Fixed in 07ef65d, and your numbers match what I measured from the tokens: `faint` is 2.25:1 on a hovered button in light and 2.23:1 in dark, so it fails everywhere, not just here. `muted` is 4.64:1 light and 4.52:1 dark on `hover`, so that is what it uses now, at `text-xs` rather than `text-2xs` — 10px was too small regardless of the ratio.

The accessible name now carries the interval at every size: each button is labelled `Good, next in 10 min`. Verified against the running app rather than by reading the code. The visible date stays a hover affordance on pointer machines, so the phone is unchanged.

Worth noting the general case: `faint` fails 4.5:1 against every surface in both themes (2.2–2.9), which is exactly why DESIGN.md says it never carries words. This was me breaking that rule, not the token being wrong.

apps/web/src/client/views/ReviewView.tsx
krambertech: Fixed in 07ef65d (now 8186b61 after a rebase). Took the second option — the copy is input-appropriate rather than input-neutral, since "tap" is wrong on a desktop and "click" is wrong on a phone.

Under `@2xl` it now reads "Reveal the card when you have it" with `<Kbd>Space</Kbd>` beside it; below that it stays "Tap the card when you have it". I verified it against the running app rather than by reading the code.

PR #40, opened 2026-09-06
feat(web): build the Insights screen

krambertech:
## ✨ What

Insights is a real screen. It answers four questions — is it sticking, am I keeping it up, what have I got, and what is coming — each as a number with the line of plain words that makes the number mean something. Under them the months, then the cards that keep coming back.

The screen was a stub promising to arrive "once there is enough history to say anything true". That sentence turned out to be the brief: every block has wording for the week you have almost no history, so the screen looks finished on day three rather than broken.

## 🤔 Why

Anki's statistics screen has fourteen graphs and people open it once. The part anyone quotes back is true retention, because Anki frames it as a comparison: you asked the scheduler for 90%, here is what you got. That framing is the whole design here. `91%` says nothing on its own, so no plate ships a number without the sentence under it, and the note is a required prop on `StatPlate` rather than an option.

Two rulings came out of it and are now written into `DESIGN.md` rather than left to drift:

**Charts draw in ink.** Amber is spent four times in the design system already. A chart wants a series colour, and if the series takes amber then Insights becomes the one screen where the accent means "data" instead of "act". Figures are `text` at graded opacity over a `plate-2` well. The exceptions are the lights, which *are* the streak, and the single bar the sentence is pointing at.

**A lit day is the fifth allowed amber.** The thirty-day strip and the month bars use the same lit glass as `SevenLights`, at two more sizes.

`RunStrip` joins consecutive days into one capsule instead of drawing them as separate marks — a row of dots has to be counted, an unbroken capsule is a run whose length you can see. Like `SevenLights` it is lit or unlit and never graded by volume: grading turns a habit picture into a scoreboard and makes a heavy Tuesday look better than a steady one.

## ✅ How to verify

- Open `/insights`. The period switch scopes the recall figure only; consistency and the months are always the whole history, because "since you started" is the only window that makes sense for them.
- `pnpm verify`, `pnpm deploy:check`, and `pnpm test:e2e:chromium` all pass on the rebased branch.

<details>
<summary>pnpm test:e2e:chromium — 12 passed</summary>

```
Running 12 tests using 1 worker
  ✓  11 [chromium] › e2e/origin-boundary.spec.ts:82:5 › clearing an installed product origin recovers from stale browser storage (1.1s)
  ✓  12 [chromium] › e2e/pwa-boundary.spec.ts:5:5 › the production product package installs its shell and starts offline (812ms)

  12 passed (58.1s)
```
</details>

Checked at 860 px and 390 px, in both rooms: all four plates settle at the same height, nothing overflows, and the smallest text on the screen measures 5.55:1 by day and 5.41:1 at night.

## 📝 Notes

Three things in here are wider than the screen and worth a look:

**The query cache had no buster.** `PersistQueryClientProvider` hydrates a week-old cache from `localStorage`. Insights is the first screen to read a nested new field, so it is the first change that could hydrate an old payload into code expecting a new one — which crashed the screen for me during development, and would have crashed it in production after this deploy. `vite.config.ts` now stamps a build id into `define` and `app-entry.tsx` passes it as `buster`. Any rebuild discards the older cache.

**Cards are counted once, at their least advanced asked direction.** My first pass read `card_states` for `recognition` only, which files every card in a production-only deck under New. It now uses the `asked` predicate from `services/decks.ts`, so a card is not Known until every direction it is asked in is.

**Consistency is grouped in SQL.** It is the one query whose cost grows with every review ever made, and all it needs is the distinct days.

Deliberately left out: an hourly breakdown, which one learner reviewing in the evening will never read, and estimated words known, which only goes up and is the closest thing on this page to XP.

The leech rows are read-only. There is no card-editing UI in the client yet — `DeckDetailView` rows only archive — so opening a word from here waits for the sheet that `CONTEXT.md` already promises from Activity.


apps/web/src/client/routes/insights.tsx
```diff
@@ -1,15 +1,15 @@
...
 
 function Insights() {
-  return (
-    <ComingSoonView
-      title="Insights"
-      body="Reviews per day, cards by state, and the words that keep coming back. Once there is enough history to say anything true."
-    />
-  );
+  const [period, setPeriod] = useState<Period>("30");
+  const { data } = useQuery(insightsQuery(Number(period) as 30 | 90 | 0));
```
krambertech: [Major] Please distinguish loading from failure here. `useQuery`'s error and refetch state is discarded, so any 500 leaves `data` undefined and `InsightsView` renders the skeleton forever. I reproduced this by forcing `/api/stats/insights` to return 500. The same `undefined` path replaces the whole screen with skeletons on each period change even though only Recall changes. Pass the query state through, render a retryable error, and retain the previous payload while the next period loads.

apps/web/src/client/lib/queries.ts
```diff
@@ -31,6 +31,13 @@ export const historyQuery = queryOptions({
   queryFn: () => api.history(7),
   staleTime: 60_000,
 });
+// Reviews change the numbers, so a cached copy is a placeholder until the refetch lands.
+export const insightsQuery = (period: 30 | 90 | 0) =>
+  queryOptions({
+    queryKey: ["insights", period],
+    queryFn: () => api.insights(period),
+    staleTime: 60_000,
```
krambertech: [Major] This query is considered fresh for 60 seconds, but none of the grade, add, archive, or deck mutations invalidate the `['insights']` prefix. Review grading currently invalidates only decks and history. If the learner reviews and opens Insights within that minute, Recall, Consistency, Collection, and Ahead can all be stale. Please invalidate Insights from every mutation that changes these aggregates, or make this query refetch on mount instead of treating the cached result as fresh.

apps/web/src/server/services/stats.ts
```diff
@@ -0,0 +1,349 @@
...
+  days: number;
+}
+
+export type Period = 30 | 90 | 0;
+
+const DAY_MS = 86_400_000;
+
+/** Local midnight for `date` shifted into the learner's day, as a UTC instant. */
+function localDayStart(at: number, tzOffset: number): number {
+  const local = new Date(at - tzOffset * 60_000);
```
krambertech: [Major] A single offset from `new Date().getTimezoneOffset()` cannot bucket historical dates in a DST zone. In Tallinn today the client sends `tz=-180`; a review at `2026-01-15T21:30Z` was local January 15 under UTC+2, but this helper applies UTC+3 and buckets it as January 16. That can fabricate or break runs and month totals. Please send an IANA timezone and resolve the offset for each timestamp, or persist the local review date, rather than applying today's offset to all history.

apps/web/src/client/components/TrendLine.tsx
```diff
@@ -0,0 +1,164 @@
...
+}) {
+  const fadeId = useId();
+  const values = points.map((p) => p.value);
+  const lo = Math.min(...values, target ?? 1) - 0.05;
+  const hi = Math.max(...values, target ?? 0) + 0.05;
+  const span = Math.max(0.14, hi - lo);
+  const floor = Math.max(0, Math.min(lo, 1 - span));
+
+  const x = (i: number) =>
+    points.length === 1 ? W / 2 : PAD_X + (i / (points.length - 1)) * (W - PAD_X * 2);
```
krambertech: [Major] The API intentionally omits empty buckets so a week away can be a gap, but `x` is derived only from the array index and `curve()` connects every adjacent item. A July point and a September point are therefore drawn as close together and as continuously as two consecutive weeks, hiding the gap and distorting elapsed time. Please derive x positions from `at`, or preserve null buckets and split the path at gaps.

apps/web/src/client/views/InsightsView.tsx
```diff
@@ -0,0 +1,291 @@
...
+              : consistency.longestRun > 1
+                ? `Longest run ${plural(consistency.longestRun, "day", "days")}. Unbroken stretches join up.`
+                : "Each block is a day. They join up when you keep going."
+          }
+        />
+
+        <StatPlate
+          label="Collection"
+          value={cards.total}
+          unit="cards"
```
krambertech: [Minor] This renders `1 cards` in the one-card state, which I confirmed in both the phone and desktop real-data render. Please make the unit singular when `cards.total === 1`.

krambertech (commented):
Changes needed. The screen itself renders cleanly with real data, but I found four correctness/state issues and one small copy defect that are worth fixing before merge.

Verified locally: `pnpm verify` passed; `pnpm deploy:check` passed; the full browser suite passed with 23 tests and the expected WebKit PWA skip; a focused real-data Insights journey passed in Chromium and WebKit at 320 px phone and 1280 px desktop widths, in dark and light themes, with all three periods exercised and no horizontal overflow. An injected 500 confirmed the indefinite loading state described inline. GitHub CI and Workers Builds are also green at this head.

apps/web/src/client/routes/insights.tsx
krambertech: Fixed in af9b958. The route now passes the query state through: `failed` renders a retryable empty state with a Try again button, and `placeholderData: keepPreviousData` keeps the previous period on screen while the next one loads, so changing the switch dims the plates rather than replacing them with skeletons.

apps/web/src/client/lib/queries.ts
krambertech: Fixed in af9b958, both ways you suggested. `insightsQuery` drops to `staleTime: 0` with `refetchOnMount: "always"`, matching how `decksQuery` already treats its persisted copy as a placeholder. Grading also invalidates the `["insights"]` prefix in `routes/review.tsx`, so a session open in another tab updates too.

apps/web/src/server/services/stats.ts
krambertech: Fixed in af9b958. The endpoint now takes an IANA zone and resolves each timestamp with `Intl.DateTimeFormat`, so your Tallinn January case lands on the 15th. The day query stays bounded: it groups by UTC day and keeps the earliest and latest review of each, since 24 hours of reviews can touch at most two local days and those two are the local dates of the first and last. Checked against the seeded data — UTC and Europe/Tallinn both resolve 105 lit days from 105 UTC groups, Pacific/Auckland resolves 106, which is the extra local day the old code could not see. Two tests cover the January/July offset difference and a month spanning a clock change.

apps/web/src/client/components/TrendLine.tsx
krambertech: Fixed in af9b958. `TrendPoint` now carries `t`, the bucket start, and x is derived from elapsed time rather than the array index. The design page mock has a two-week gap in it so the case stays visible.

apps/web/src/client/views/InsightsView.tsx
krambertech: Fixed in af9b958.

krambertech (commented):
Additional UI pass: the visual craft is strong across the rendered desktop, phone, light, and dark states. I found one phone discoverability issue plus a few interaction and copy details that keep it from feeling finished everywhere.

apps/web/src/client/views/Shell.tsx
```diff
@@ -25,19 +25,18 @@ export interface NavDeck {
...
   to: string;
   label: string;
   icon: LucideIcon;
   exact?: boolean;
-  later?: boolean;
 }[] = [
   { to: "/today", label: "Today", icon: Sun, exact: true },
   { to: "/library", label: "Library", icon: BookMarked },
-  { to: "/insights", label: "Insights", icon: ChartNoAxesColumn, later: true },
+  { to: "/insights", label: "Insights", icon: ChartNoAxesColumn },
```
krambertech, line 39: [Major] This adds Insights to the desktop sidebar, but the phone shell has no route to it. The rendered phone UI still exposes only Today and Library in `PillNav`, and there is no other `/insights` link in the client, so a phone learner can reach the screen only by typing its URL. Please add a deliberate phone entry point, for example from Today or the profile surface, while keeping the two-item pill decision intact.

apps/web/src/client/views/InsightsView.tsx
```diff
@@ -0,0 +1,323 @@
...
+  return (
+    <Page>
+      <PageHeader
+        title="Insights"
+        sub={
+          consistency.daysAllTime > 0
+            ? `${plural(consistency.daysAllTime, "day", "days")} since your first review`
+            : undefined
+        }
+        actions={switcher}
```
krambertech: [Major] Placing this unlabeled switcher in the page header makes 30/90/All read as a filter for the whole Insights screen, but changing it updates only Recall; Consistency, Collection, Ahead, Month by month, and Keeps coming back stay unchanged. The recall-only explanation is screen-reader-only, so sighted learners get no scope cue. Please move the control into the Recall plate or add a visible label such as "Recall period" so the interaction matches its effect.

apps/web/src/client/views/InsightsView.tsx
```diff
@@ -0,0 +1,323 @@
...
+ * the cards that keep coming back.
+ *
+ * Every figure draws in ink except the lights, which stay amber because they are the
+ * streak's lights. A chart series in amber would make this the one screen where the accent
+ * means "data" rather than "act".
+ */
+export function InsightsView({ data, period, onPeriod, failed, busy, onRetry }: InsightsProps) {
+  const switcher = (
+    <Segmented
+      size="sm"
```
krambertech, line 61: [Minor] The small segmented buttons are 28 px high with a 6 px pseudo-element inset on each side, giving an effective 40 px vertical target. That misses PRODUCT.md's 44 px minimum for phone tap targets on a primary interaction shown at the top of this screen. Please use the 44 px treatment on touch layouts while retaining the compact desktop size if needed.

apps/web/src/client/views/InsightsView.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const nothingYet = consistency.daysAllTime === 0 && cards.total === 0;
+
+  if (nothingYet) {
+    return (
+      <Page>
+        <PageHeader title="Insights" />
+        <EmptyState
+          lantern="unlit"
+          title="Nothing to say yet"
+          body="Reviews per day, how much is sticking, and the words that keep coming back. This fills in once there is some history behind you."
```
krambertech: [Minor] "The words that keep coming back" narrows this empty state to vocabulary, while Lymi cards can also hold phrases, terms, and concepts and CONTEXT.md explicitly avoids "word" as the product noun. "The cards that keep coming back" would stay accurate across the product's supported material.

apps/web/src/client/views/InsightsView.tsx
```diff
@@ -0,0 +1,323 @@
...
+                className="flex items-center justify-between gap-4 border-t border-edge px-5 py-3"
+              >
+                <span className="min-w-0">
+                  <span className="font-medium text-text" lang={c.language ?? undefined}>
+                    {c.term}
+                  </span>
+                  {c.meaning && <span className="ml-2 text-sm text-muted">{c.meaning}</span>}
+                </span>
+                <span className="shrink-0 text-sm text-muted tabular-nums">
+                  {c.lapses} of {c.reviews}
```
krambertech: [Minor] A row ending in "7 of 12" does not say what either number counts. Even with the section heading, the relationship remains guesswork. Please include the units in the visible or accessible text, for example "7 forgotten of 12 reviews", so each number carries its meaning.

apps/web/src/client/views/Shell.tsx
krambertech: Fixed in 7af2f00. Insights now sits in You's Oversight group beside Activity and Archived, which is already how the phone reaches those two. The pill stays two items wide, so ADR 0005 is intact.

apps/web/src/client/views/InsightsView.tsx
krambertech: Fixed in 7af2f00 by moving it, which I think is the better of the two options you offered: a label would explain the mismatch where the position causes it. The switcher is now on the Recall plate's label row. StatPlate reserves that row at a fixed height whether or not a plate has a control, so all four plates still measure 243 px and the grid stays aligned.

apps/web/src/client/views/InsightsView.tsx
krambertech: Fixed in 7af2f00. The `sm` pseudo-element inset goes from 6 px to 8 px either side of the 28 px button, so the target measures 44 px — I checked the computed `::before` in the rendered phone frames. It is in `Segmented` rather than at this call site, so every small segmented control in the app gets it.

apps/web/src/client/views/InsightsView.tsx
krambertech: Fixed in 7af2f00 — "the cards that keep coming back". You are right that CONTEXT.md rules out "word" as the product noun; I also swept the rest of the screen and the service for it.

apps/web/src/client/views/InsightsView.tsx
krambertech: Fixed in 7af2f00. The row now reads "7 forgotten · 12 reviews", so each number carries its unit. Checked for overflow at 390 px: none, and the term column still holds its meaning.

PR #41, opened 2026-09-12
feat(web): rebuild Today around one question and one action

krambertech:
## ✨ What

Today asks one question and offers one action: the due count is the heading, the lantern is the only object on the screen, and the decks holding the due cards are named under it, each name opening that deck. Nothing due is no longer a dead end — the lantern goes out and Add a word becomes the primary action.

The streak becomes something worth looking at: the flame and the run on one line, the week under it at roughly triple the old area, each day's light stepping through three shades of amber by how full that day was.

## ✅ How to verify

- Sign in, open `/today` with cards due: count as the heading, deck names as links, one Review button, the week under it. Grade everything and come back: unlit lantern, "Nothing due", Add a word.
- Switch the OS to dark: an empty day and a light day in the week must read as different cells. That was the bug the first pass of the shading had.
- `pnpm verify` — green (Biome, build, typecheck, 21 + 3 + 70 unit tests). Full `pnpm test:e2e` — 23 passed, 1 skipped (`pwa-boundary` skips itself outside Chromium) across Chromium and WebKit in one run.

<details><summary>e2e tail</summary>

```
  ✓  20 [webkit] › e2e/origin-boundary.spec.ts:5:5 › the public surface has no install contract while the product keeps its PWA (1.5s)
  ✓  21 [webkit] › e2e/origin-boundary.spec.ts:34:5 › each origin exposes only its own route and indexing contract (427ms)
  ✓  22 [webkit] › e2e/origin-boundary.spec.ts:55:5 › the public Worker owns beta signup without exposing product APIs (605ms)
  ✓  23 [webkit] › e2e/origin-boundary.spec.ts:82:5 › clearing an installed product origin recovers from stale browser storage (1.4s)
  -  24 [webkit] › e2e/pwa-boundary.spec.ts:5:5 › the production product package installs its shell and starts offline

  1 skipped
  23 passed (2.2m)
```

</details>

## 🤔 Why

The old screen was a launcher wearing a dashboard's stat card: the streak plate took a third of the primary row and said one fact four ways (12 days, best 26, seven lights, "reviewed 7 of the last 7"). PRODUCT.md names the home-screen dashboard as an anti-reference, so the plate is gone and `StreakPlate` and `StreakPill` with it. The page also knew less than the rail — the sidebar listed decks with due counts while the hero said "Across 2 decks" — and the state the screen is in every evening after a session offered nothing to do.

Grading the week by volume is a deliberate exception to what `RunStrip` on Insights refuses to do. Over thirty days, shading by volume makes a habit picture into a scoreboard; over seven it is "I turned up" versus "I did the lot", which the learner already knows. DESIGN.md now says so, so the two components do not read as an accident. The steps mix toward `--glass` rather than fading to transparency: a translucent amber on the dark canvas lands at the same lightness as an unlit day.

## 📸 Screenshots

Captured from the running app on a seeded local account. Before is `origin/main` at b2c82fc. Full set, both themes: https://share.kateryna.codes/lymi-pr-today

**Cards due — iPhone**

| Before | After |
|---|---|
| Before: cards due, phone, light | After: cards due, phone, light |
| Before: cards due, phone, dark | After: cards due, phone, dark |

**Cards due — desktop**

| Before | After |
|---|---|
| Before: cards due, desktop, light | After: cards due, desktop, light |
| Before: cards due, desktop, dark | After: cards due, desktop, dark |

**Nothing due** — the state that previously offered no action

| Before | After |
|---|---|
| Before: nothing due, phone | After: nothing due, phone |
| Before: nothing due, desktop | After: nothing due, desktop |

**First run**

| Before | After |
|---|---|
| Before: first run, phone | After: first run, phone |

## 📝 Notes

Today still shows no words. "Cards you added since your last lesson" and "cards that became Known this week" are the two candidates, each one small query; they are deliberately not in this PR. The arrivals section and the worded forecast keep their props on the view but nothing supplies them yet — the Insights payload now carries a seven-day forecast, so that half is one wiring change away.

`e2e/auth.ts` asserted a heading named "Today" after sign-in; Today no longer has one, so it now asserts the hero heading in either of its states.

Self-review caught that the history query fetched seven days, so the run could never read past "7 days in a row". Review then caught that fetching ninety only moved the cap. The history response now carries `streak`, counted on the server with no window, and `e2e/streak-exact.spec.ts` seeds nine consecutive days and asserts 9 from the API and on the screen. Review also asked for the day levels to survive without colour and to reach assistive technology: the light now climbs the glass as well as brightening, and the per-day counts are in the image's description. The hero waits for both queries so no placeholder can leave after it lands. The "after" screenshots above are from the final head.


apps/web/src/client/components/SevenLights.tsx
```diff
@@ -3,46 +3,95 @@ import { clsx } from "clsx";
...
   const lit = days.filter((n) => n > 0).length;
+  // A floor under the reference, so a week of two-card days does not read as a full week.
+  const reference = Math.max(...days, 10);
+  const large = size === "lg";
   return (
     <div
-      className={clsx("inline-flex items-end gap-2", className)}
+      className={clsx("inline-flex items-end", large ? "gap-2.5" : "gap-2", className)}
       role="img"
       aria-label={`Reviewed on ${lit} of the last ${days.length} days`}
```
krambertech: [P2] Expose activity levels without relying on color

The new three-level activity signal is encoded only by fill color. Because this wrapper is an atomic role=img, assistive technology receives only the summary count, while the per-day title values are unavailable to keyboard and touch users. Please include the daily counts in the accessible description and add a non-color visual distinction for the levels.

apps/web/src/client/lib/queries.ts
```diff
@@ -26,9 +26,13 @@ export const queueQuery = (deckId?: string) =>
...
   });
+/**
+ * Ninety days, the server's ceiling, because the streak is counted from this array and a
+ * seven-day window would cap every run at seven. The lights slice the last seven themselves.
+ */
 export const historyQuery = queryOptions({
-  queryKey: ["history", 7],
-  queryFn: () => api.history(7),
+  queryKey: ["history", 90],
+  queryFn: () => api.history(90),
```
krambertech: [P2] Do not cap the displayed streak at 90 days

This fixes the previous seven-day cap but still makes streakLength see at most 90 days, so a 91-day or longer streak is reported as 90. Please return the exact current streak separately, or provide enough continuation information to determine whether the run extends beyond this window.

apps/web/src/client/views/TodayView.tsx
```diff
@@ -76,99 +82,114 @@ export function TodayView({
...
-          )}
-        </section>
+        )}
 
-        <StreakPlate days={history} className="hidden @3xl:flex" />
-      </div>
+        {/* The streak sits outside the branch above: it arrives on its own query, after the
+            decks, and the hero is centred, so it has to hold its room in both states or the
+            whole screen jolts upward when the lights land. */}
+        {loading || history === undefined ? (
```
krambertech: [P3] Preserve the first-run skeleton footprint

If decks resolve before an empty history response, this skeleton is rendered and then removed entirely because everReviewed is false. Since the whole section is vertically centered, the hero visibly recenters after loading. Please keep a stable footprint or wait for both queries before rendering the resolved hero.

apps/web/src/client/components/SevenLights.tsx
krambertech: Done in HEAD.

The per-day counts now live in the image's own description — `Reviewed on 5 of the last 7 days: Mon 12, Tue 3, Wed none, …` — so an atomic `role="img"` reads everything a sighted learner sees, and nothing depends on a `title` a keyboard or touch user cannot reach.

The non-colour cue is height: the glass is always drawn, and the light inside it climbs to a third, two thirds, or the top, in step with the colour. Same information twice, which is what PRODUCT.md's "status is never colour alone" asks for.

apps/web/src/client/lib/queries.ts
krambertech: Done in HEAD — you were right that this fix only moved the cap.

The history response now carries `streak`, counted on the server with no window: reviews grouped by UTC day and resolved to the learner's zone the way `stats.ts` already does for Insights, so the query is bounded by days rather than by reviews and the run and the lights can never disagree. Today reads `streak` and derives from `history` only where no server value exists (the design page).

The lights are back to seven days, since that is all they draw. Evidence: `e2e/streak-exact.spec.ts` seeds nine consecutive backdated days and asserts `streak === 9` from the API and "9 days in a row" on the screen — a run counted from the seven-day window would say 7.

apps/web/src/client/views/TodayView.tsx
krambertech: Done in HEAD, the second way you suggested: one loading gate for both queries.

The resolved hero now waits for decks *and* history, and the loading skeleton reserves the streak block's footprint. There is no longer a state in which a placeholder can leave after the hero has landed, on a first run or otherwise. Both queries fire in parallel and history is seven integers, and on a returning device the persisted cache paints both before either request returns, so the wait is not observable in practice.

PR #42, opened 2026-09-12
feat(web): redesign Library and open a word with its whole history

krambertech:
## ✨ What

Library is every deck as a card with a face: its name and full language name, a stripe in ink that says how its cards split between known, learning and new, and one amber line for what is due. A card opens as a page with everything Lymi knows about it and its whole history, read first and edited on request, and on desktop that page sits in a panel beside the deck's list.

The deck's list is plain rows grouped by the lesson each card came from, with state as the filter above the list rather than a pill on the row. Its final shape is still open; this PR carries the version needed to open a word.

## ✅ How to verify

- `pnpm verify` — format, lint, build, typecheck, 113 tests across core, site and web: green.
- `pnpm test:e2e:chromium` — 15 passed. Two specs moved from table cells to the list rows, and `e2e/word-detail.spec.ts` opens a word, edits a field, moves it and archives it.
- `pnpm deploy:check` — dry run: green.
- Open a deck, press a row: the word opens beside the list (desktop) or as its own screen (phone). Press a field's text or choose Edit from the word's menu, change it, blur: it saves, and History gains an "Edited" line. Move to… opens a sheet of the other decks. `J`/`K` walk the list, `Esc` closes.

<details><summary>pnpm verify</summary>

```
packages/core test:       Tests  21 passed (21)
apps/site test:       Tests  3 passed (3)
apps/web test:       Tests  89 passed (89)
```
</details>

<details><summary>pnpm test:e2e:chromium</summary>

```
  15 passed (1.5m)
```
</details>

## 🤔 Why

The earlier draft on this branch never decided what Library is for: due was said four times on one screen, every deck carried a green progress bar, and every row ended in a status pill and an AI badge. This version says each fact once. Progress is the deck's shape, not a score, drawn the way Insights draws everything: ink at three opacities. Due is the only amber. The one number FSRS can honestly give, retrievability averaged over a deck, is a sentence in the header rather than a ring. Language is a word, never a code.

A word's history needs two reads the client did not have: `GET /api/cards/:id/history` returns its reviews and its audit writes, newest first, with schemas in `packages/core` so it lands in the OpenAPI document. Editing goes through the existing `PATCH /api/cards/:id`, which the client now calls.

## 📸 Screenshots

Captured from the seeded local app, desktop and iPhone, both rooms. Attached in the review thread; the CLI cannot place images in a PR body.

## 📝 Notes

- The previous head of this branch (ae38cb2) also floated the Combobox through the Popover API and let the capture sheet add several words in a row. Both are out of this PR's scope now and can come back on their own.
- Not here, deliberately: an Archived screen and an archived count in Library (no endpoint yet), "Start over" for a card's schedule (no reset endpoint), and asking the AI again for a field (no enrich endpoint). The interface carries no control for any of them.
- The panel and the Move sheet are new dialog and popover surfaces, so the WebKit project matters in CI.


apps/web/src/client/views/WordView.tsx
```diff
@@ -0,0 +1,564 @@
...
+    const refs = { meaning: meaningRef, example: exampleRef, notes: notesRef };
+    const el = refs[focusRef.current].current;
+    el?.focus();
+    if (el) {
+      const n = el.value.length;
+      el.setSelectionRange(n, n);
+    }
+    focusRef.current = null;
+  }, [editing]);
+  const commit = (key: keyof WordPatch, before: string | null) => (v: string) => {
```
krambertech: [P1] Editing meaning or example sends only the changed text, so the previous `ai` or `lesson` source remains. I verified that a learner-edited meaning is still persisted and displayed as "The AI wrote this"; please include the corresponding source as `manual` whenever the learner changes these fields.

apps/web/src/client/views/DeckDetailView.tsx
```diff
@@ -109,79 +189,195 @@ export function DeckDetailView({
...
-    <>
-      <ChevronLeft className="size-4" aria-hidden="true" />
-      Library
-    </>
+
+  const word = open && deck && (
+    <WordView
+      key={open.card.id}
+      card={open.card}
+      state={open.state}
```
krambertech, line 266: [P1] The detail receives only the deck list's leading scheduling state. A card configured for both directions therefore shows recognition only, while its recognition and production reviews appear as indistinguishable rows; please fetch and render every asked direction with direction-labelled history.

apps/web/src/client/views/WordView.tsx
```diff
@@ -0,0 +1,564 @@
...
+        <span className="text-sm font-medium text-text-2">{label}</span>
+        {aside && <span className="text-xs text-muted">{aside}</span>}
+      </div>
+      <button
+        type="button"
+        onClick={onEdit}
+        aria-disabled={!onEdit}
+        aria-label={`Edit ${label.toLowerCase()}`}
+        className={clsx(
+          "-mx-2 rounded-sm px-2 py-1 text-left text-md leading-relaxed transition-colors",
```
krambertech: [P1] A single-line editable field measures about 33px high, below Lymi's 44px touch-target requirement. Please give the button at least `min-h-11` while preserving its current visual treatment.

apps/web/src/client/views/WordView.tsx
```diff
@@ -0,0 +1,564 @@
...
+        <h1 className="flex min-w-0 items-center gap-3 text-3xl font-medium leading-[1.05] tracking-[-0.03em]">
+          <span className="min-w-0 break-words" lang={card.language ?? undefined}>
+            {card.term}
+          </span>
+          <IconButton
+            label={`Say ${card.term}`}
+            variant="secondary"
+            round
+            size="sm"
+            onClick={onPlayAudio}
```
krambertech: [P2] The production route never provides `onPlayAudio`, so every word displays an aria-disabled Say button, including cards without a language. Please wire this through `api.audioUrl` as Review does, and render it only when audio is available.

apps/web/src/client/views/WordView.tsx
```diff
@@ -0,0 +1,564 @@
...
+                {[...reviews]
+                  .reverse()
+                  .slice(-12)
+                  .map((r) => (
+                    <GradeMark key={r.id} rating={r.rating} />
+                  ))}
+              </span>
+            )}
+          </div>
+          <ol className="grid">
```
krambertech, line 569: [P2] Reviews are rendered first and audit events second, so a recent edit can appear below older reviews despite the newest-first promise. Please merge the collections and sort by timestamp, retaining the audit event ID for a collision-safe React key.

apps/web/src/client/routes/library.$deckId.tsx
```diff
@@ -20,14 +26,26 @@ function Deck() {
...
   const [undo, setUndo] = useState<{ id: string; term: string } | null>(null);
 
+  const events = useMemo(() => history.data?.events.map(describeEvent), [history.data]);
+
+  const setOpen = (id: string | null) =>
+    navigate({
+      to: "/library/$deckId",
+      params: { deckId },
+      search: id ? { card: id } : {},
+      replace: !!openCardId && !!id,
```
krambertech: [P2] Clearing the card query pushes another deck-list entry because `replace` is false when `id` is null. I verified that closing the panel and pressing browser Back reopens the card; please close by replacing or popping the card entry.

apps/web/src/client/routes/library.$deckId.tsx
```diff
@@ -47,6 +65,14 @@ function DeckPage() {
       invalidate();
     },
   });
+  const save = useMutation({
```
krambertech, line 81: [P2] This mutation has no error handling while the editor closes immediately on blur or Done. A rejected or offline PATCH leaves the learner without an explanation or retry path; please preserve the draft and surface a recoverable error.

krambertech (commented):
I found three ship-blocking issues and four should-fix regressions in the new word-detail flow. The current checks are green, but the existing E2E suite does not exercise opening, editing, moving, archiving, or reading history on this screen.

apps/web/src/client/views/WordView.tsx
krambertech: Fixed in 75bc97d. A change to the meaning or example now sends `meaningSource` / `exampleSource: "manual"` with the text, so the label reads "You wrote this". Covered by e2e/word-detail.spec.ts.

apps/web/src/client/views/DeckDetailView.tsx
krambertech: Fixed in 75bc97d. `GET /api/cards/:id/history` now returns `states`, one per direction. The page draws a "Right now" block per direction and prefixes each review line with its direction when the card is asked both ways.

apps/web/src/client/views/WordView.tsx
krambertech: Fixed in 75bc97d. The read-only field button is `min-h-11` with the same visual treatment.

apps/web/src/client/views/WordView.tsx
krambertech: Fixed in 75bc97d. The route plays through `api.audioUrl` the way Review does, the button renders only when the card has a language, and a failed play shows the same toast Review uses.

apps/web/src/client/views/WordView.tsx
krambertech: Fixed in 75bc97d. Reviews and audit writes merge into one timeline sorted by timestamp, newest first, keyed by `r-<review id>` or `e-<audit id>`.

apps/web/src/client/routes/library.$deckId.tsx
krambertech: Fixed in 75bc97d. Only opening from the list pushes an entry; walking and closing replace it, so Back after a close returns to Library rather than reopening the word.

apps/web/src/client/routes/library.$deckId.tsx
krambertech: Fixed in 75bc97d. A rejected PATCH keeps the patch and shows "Couldn't save" with a Retry action; success clears it.

PR #43, opened 2026-09-12
feat(mcp): give the MCP server its tools

krambertech:
## ✨ What

A connected assistant can now work with the learner's decks: sixteen MCP tools over the same service layer as the REST routes, so Claude, Codex or ChatGPT can list, create, edit, archive and restore decks; search, get, add, update, archive and restore cards; and read due counts, settings and Insights. Card search is new and is exposed on REST too, as `GET /api/cards`, so the two surfaces stay level. Review grading stays learner-only on both.

## ✅ How to verify

- `pnpm --filter @lymi/web test src/server/mcp` runs the in-memory MCP client against every tool, the read-only refusal, and the HTTP transport with synthetic claims.
- Verified end to end on the local dev server with a pre-registered native client: OAuth sign-in, the consent screen, token exchange with PKCE, then every tool called with the real bearer token. Each write landed in `audit_log` with actor `mcp`.
- `pnpm verify`, `pnpm deploy:check:product` and `pnpm test:e2e:chromium` pass.

<details>
<summary>Tool calls with the real token</summary>

```
tools/list  → list_decks, get_deck, search_cards, get_card, due_counts, add_cards, update_card, archive_card, restore_card, create_deck, update_deck, archive_deck, restore_deck, get_settings, update_settings, get_insights
add_cards   → {"added": 2, "skipped": 1, ...}   one duplicate inside the batch, reported with the existing card
update_card → pronunciation and example set, exampleSource "ai"
archive_card / search_cards archived=true / restore_card → ok
create_deck / update_deck / archive_deck / restore_deck → ok
get_settings → {"meaningLanguage": "en"}; update_settings round trip uk → en
get_insights period=30 timezone=Europe/Tallinn → recall, consistency, forecast, leeches
get_card cardId=nope → tool error "Card not found"
add_cards cards=[] → input validation error before any service runs
audit_log where actor='mcp' → create, update, archive, restore for card and deck
```

</details>

## 🤔 Why

A meaning or example an assistant sends without naming its source is stored as `ai`. The label exists so the learner never mistakes an assistant's guess for what the lesson said, and the tool copy tells the assistant to say `lesson` or `ai` itself. `create_deck` is in the set so a lesson in a new language does not send the learner to the app mid-conversation.

A write on a read-only token is refused with a message that says to reconnect with write ticked, rather than an HTTP 403, because over MCP the assistant reads the tool result and relays it.

## 📝 Notes

- `enrich` is not in the set. The enrichment service from step 7 of the integrations plan does not exist yet; the assistant writes meanings at add time until it does.
- The Activity screen is still a stub. Cards from an assistant show in Library and the deck views and carry `createdBy: mcp`.
- The docs at `/docs/mcp` no longer say the tools are missing, and list the sixteen.


apps/web/src/server/mcp/server.ts
```diff
@@ -8,21 +39,537 @@ export interface McpPrincipal {
...
+      description:
+        "Change fields on one card. Send only what changes; a field left out keeps its text. Setting deckId moves the card. Needs write.",
+      inputSchema: z.object({ cardId: z.string().min(1) }).extend(CardPatch.shape),
+      outputSchema: CardOut,
+      annotations: { ...write, idempotentHint: true },
+    },
+    ({ cardId, ...patch }) =>
+      run(async () => {
+        denyReads(principal);
+        return result(cardOut(await updateCard(ctx, cardId, patch)));
```
krambertech: **[P1] Preserve honest provenance on card edits**

`update_card` forwards `CardPatch` directly, unlike `add_cards`. Changing a meaning or example without its source therefore preserves the previous source or leaves it null, allowing model-written text to remain labelled `lesson` or `manual`. Please apply MCP-specific source defaults to edits and reject `manual` as an MCP-supplied source.

apps/web/src/server/services/cards.ts
```diff
@@ -171,6 +182,51 @@ function dupKey(language: string | null, normalizedTerm: string): string {
...
+  const query = search.query?.trim();
+  const pattern = query ? `%${escapeLike(normaliseTerm(query))}%` : null;
+  return db
+    .select({ card: schema.cards, deckName: schema.decks.name })
+    .from(schema.cards)
+    .innerJoin(schema.decks, eq(schema.decks.id, schema.cards.deckId))
+    .where(
+      and(
+        eq(schema.cards.userId, userId),
+        search.archived ? isNotNull(schema.cards.archivedAt) : isNull(schema.cards.archivedAt),
```
krambertech: **[P2] Exclude archived decks from active card search**

The search joins decks but filters only `cards.archivedAt`. Because archiving a deck leaves its cards unarchived, those hidden cards still appear in default REST and MCP searches. Active results should require both the card and deck to be unarchived; archived search semantics should cover both archive paths.

apps/web/src/server/services/cards.ts
```diff
@@ -171,6 +182,51 @@ function dupKey(language: string | null, normalizedTerm: string): string {
...
+        search.deckId ? eq(schema.cards.deckId, search.deckId) : undefined,
+        search.language ? eq(schema.cards.language, search.language) : undefined,
+        pattern
+          ? or(
+              contains(schema.cards.normalizedTerm, pattern),
+              contains(sql`lower(${schema.cards.meaning})`, pattern),
+              contains(sql`lower(${schema.cards.example})`, pattern),
+              contains(sql`lower(${schema.cards.notes})`, pattern),
```
krambertech: **[P2] Make multilingual text search case-insensitive**

The query is Unicode-lowercased, but SQLite `lower()` only folds ASCII. For example, stored `Привіт` does not match a search for `привіт`; accented uppercase examples fail similarly. Please use persisted normalized search values or Unicode-aware indexing for meaning, example, and notes.

krambertech (commented):
Three correctness findings from a local review. The full verification gate, deployment dry run, focused MCP tests, and Chromium E2E all pass; these cases are not covered by the current tests.

apps/web/src/server/mcp/server.ts
krambertech: Fixed in c77b5d8. `update_card` now runs the same `withAiSourceDefaults` as `add_cards`: a changed meaning or example without a stated source is stored as `ai`. Both tools now accept only `lesson` or `ai` as a source; `manual` is rejected at the schema. Covered in server.test.ts.

apps/web/src/server/services/cards.ts
krambertech: Fixed in c77b5d8. Active search requires both `cards.archivedAt` and `decks.archivedAt` to be null; `archived=true` returns cards hidden by either path. Checked on the dev server: cards of an archived deck vanish from the default search and appear under archived.

apps/web/src/server/services/cards.ts
krambertech: Fixed in c77b5d8. The text match moved out of SQL: the query runs the deck, language and archived filters, then the rows are folded with the duplicate-key rule (`normaliseTerm`, NFC + toLocaleLowerCase) and matched in memory, capped at 5000 scanned rows. A persisted folded column is the upgrade if a collection outgrows that. Checked on the dev server: `JÄLKEEN` finds `jälkeen`, which SQLite's `lower()` could not. Unit tests cover Cyrillic and accented cases.

PR #44, opened 2026-09-12
docs(product): decide shared decks and plan the build

krambertech:
## Summary

A learner can share a deck with a class: [ADR 0011](docs/adr/0011-a-shared-deck-is-one-deck-with-many-learners.md) decides that a shared deck is one live deck with many learners, that each learner's schedule is their own, and that a deck's join link or named invitation is how a classmate gets into Lymi. [The plan](docs/plans/2026-09-12-shared-decks.md) orders six steps into four pull requests, and CONTEXT.md gains the sharing vocabulary.

## Verification

- `pnpm verify` passes. Docs only.
- The plan's join flow was checked against the installed code: Better Auth 1.7.2 passes the request context to both hooks, and sign-in only returns to the paths in `PRODUCT_PATHS`, so the join completes in a `session.create.after` hook rather than on the join page.

## Follow-ups

Trying a deck before joining, without an account, is recorded in the proposal as a later slice.


docs/adr/0011-a-shared-deck-is-one-deck-with-many-learners.md
```diff
@@ -0,0 +1,58 @@
...
+
+The owner sees who is a member and when they joined. The owner sees nothing about a member's reviews.
+
+## The deck's invitation admits the account
+
+Sign-in is Google only. The `user.create.before` hook in `apps/web/src/server/auth.ts` refuses any email not in `ALLOWED_EMAILS`. That gate stays. A second gate opens next to it. An email with an unrevoked named invitation may create an account. So may anyone who reaches sign-in from a valid join link.
+
+A deck has one join link. The owner can turn it off. A deck can also have named invitations, one per Google email. Lymi sends no email; the owner passes the link on.
+
+Turning the link off removes nobody. Removing a member takes the deck out of their Library. Their states and reviews stay, so re-adding them resumes where they were.
```
krambertech: Removing a member is not durable while the generic join link stays active: the removed account still has the link and can immediately join again. Please define whether owner removal creates a block/tombstone (distinct from voluntary leave), enforce it in `join`, and cover rejoining in the tests.

docs/plans/2026-09-12-shared-decks.md
```diff
@@ -0,0 +1,103 @@
...
+| 2. Front door | 2, 3 | Hook tests. A fetch of `/join/<token>` shows the Open Graph tags. Deploy and try with one classmate. |
+| 3. Sharing in the app | 4, 6 | The two-context Playwright test. `pnpm test:e2e`, both browsers. |
+| 4. Integrations | 5 | Unit tests for 403 through a key and a token. OpenAPI snapshot updated. |
+
+PR 4 can follow PR 1 in any order.
+
+## Constraints checked against the installed code
+
+- Better Auth 1.7.2: `user.create.before(user, context)` and `session.create.after(session, context)` receive a `GenericEndpointContext | null`, so the cookie is readable in both. Treat null as "no link".
+- The Worker serves the SPA from `app.notFound` in `apps/web/src/server/index.ts`, so `/join/:token` needs no ordering trick.
```
krambertech: `decideOriginRoute` runs before Hono and currently only special-cases the exact `/join` path; `/join/:token` falls through to 404. Please include the origin-routing change and test in step 3, and remove the claim that no routing work is needed.

docs/adr/0011-a-shared-deck-is-one-deck-with-many-learners.md
```diff
@@ -0,0 +1,58 @@
+---
+status: proposed
+date: 2026-09-12
```
krambertech, line 3: This status conflicts with the rest of the patch: the proposal says accepted and the plan says the decision is made, while current `docs/adr/README.md` defines `proposed` as approval pending. Please either keep the proposal exploratory and remove the plan, or mark this accepted and update `PRODUCT.md`, which still defines the first version as private and single-user.

docs/plans/2026-09-12-shared-decks.md
```diff
@@ -0,0 +1,103 @@
...
+  A-->>C: session cookie, redirect to /library/:deckId
+```
+
+The join completes in the session hook because `safeProductReturnPath` in `apps/web/src/shared/origins.ts` only returns sign-in to the paths in `PRODUCT_PATHS`, and `/join` is not one. A hook also runs once per sign-in whatever the browser does with the callback page. A signed-in learner who opens the link gets the same page, and **Join** calls `join` directly.
+
+## Order of work
+
+Each step leaves the app deployable.
+
+1. **Membership and read paths.** One migration: `deck_members` and the `card_states` index swap. `deckAccess` and `visibleDecks` in `services/decks.ts`. Route `listDecks`, `getDeck`, `listDeckCards`, `reviewQueue`, `reviewHistory`, the stats queries, and the reminder query in `push-delivery.ts` through membership, filtering states on the caller's `user_id`. Writes keep the owner filter. Adding a card creates states for every member. New `services/members.ts` with `join` and `leave`; `join` creates states the way `backfillDirection` does. Tests: a member sees the deck and its cards, grades their own state, cannot write a card; the owner's queue is unchanged.
```
krambertech: The membership read-path inventory misses `getCard` and `searchCards`; `pronunciationAudio` also calls `getCard`, so a member would get 404 for audio and MCP `get_card` and search would omit shared cards. Please add these paths, decide how the audio cache write is authorized, and test member access.

docs/adr/0011-a-shared-deck-is-one-deck-with-many-learners.md
```diff
@@ -0,0 +1,58 @@
...
+- **Open sign-up.** Rejected. It admits people no deck invited.
+- **Keep `ALLOWED_EMAILS` and add classmates by hand.** Rejected. Every invitation would edit a production secret.
+- **Render the join page on `lymi.app`.** Rejected. Sign-in and sessions live only on `my.lymi.app` under [ADR 0008](0008-public-website-and-product-use-separate-origins.md).
+- **A `shared_decks` table next to `decks`.** Rejected. One table lets Library, the queue, and the API treat a shared deck as a deck.
+- **Show the owner how many cards each member knows.** Rejected. That is a classroom product with consent questions this decision does not take on.
+
+## Consequences
+
+- Reads go through membership instead of `decks.user_id` or `cards.user_id`: `listDecks`, `getDeck`, `listDeckCards`, `reviewQueue`, the due counts, and the reminder query in `push-delivery.ts`. Writes keep the owner filter.
+- Joining creates states for every active card and asked direction, due now, as [ADR 0007](0007-a-decks-direction-is-a-filter-not-a-migration.md) does when a direction turns on. Adding a card creates states for every member.
```
krambertech: State fan-out also needs to cover direction changes, not only joins and new cards. Today `openDirections` creates missing states for the owner and `updateCard` does not backfill directions, so existing members would never receive a newly enabled review direction; add both deck- and card-direction transitions to the plan and tests.

docs/plans/2026-09-12-shared-decks.md
```diff
@@ -0,0 +1,103 @@
...
+## Pull requests
+
+| PR | Steps | Gate |
+| --- | --- | --- |
+| 1. Membership | 1 | Unit tests for read paths and write refusals. `pnpm test:e2e:chromium` for the owner's flows. |
+| 2. Front door | 2, 3 | Hook tests. A fetch of `/join/<token>` shows the Open Graph tags. Deploy and try with one classmate. |
+| 3. Sharing in the app | 4, 6 | The two-context Playwright test. `pnpm test:e2e`, both browsers. |
+| 4. Integrations | 5 | Unit tests for 403 through a key and a token. OpenAPI snapshot updated. |
+
+PR 4 can follow PR 1 in any order.
```
krambertech: PR 3 cannot precede PR 4 as written: the Sharing UI needs the learner role, and Library also needs an owner display field, while current deck summaries expose neither. Move the required response contract earlier or make PR 3 depend on the integrations and contract PR.

docs/plans/2026-09-12-shared-decks.md
```diff
@@ -0,0 +1,103 @@
...
+2. **Invitations and the front door.** `deck_invitations`. The `user.create.before` hook allows an email in `ALLOWED_EMAILS`, an unrevoked email invitation, or a `lymi_join` cookie naming an unrevoked link. A `session.create.after` hook completes the join. Services: `createLink`, `revokeLink`, `inviteEmail`, `revokeInvitation`, `listMembers`, `removeMember`, with API routes. Tests: each gate, a revoked token, and a repeated callback joining once.
+3. **The join page.** `GET /join/:token` next to the API routes, returning server HTML with the Lymi tokens inlined and `og:title`, `og:description`, `og:image`. It shows the deck's name, card count, owner, and first few terms. Three states: live, turned off, deck archived. Signed out shows **Join with Google**; signed in shows **Join**, or **Open deck** for a member. `POST /join/:token` sets the cookie and redirects. Load `impeccable` and `ux-writing` first. Check the preview by fetching the route and by pasting the link into a chat.
```
krambertech: The unauthenticated join page adds a public read of actual card terms, although ADR 0011 only decides to show deck name, count, and owner and the proposal defers trying cards before joining. Given the private-card invariant, remove the terms from this slice or make that privacy boundary an explicit accepted decision with appropriate safeguards.

PR #49, opened 2026-09-12
feat(web): localize the product's plumbing with Lingui

krambertech:
## Summary

Phase 1 of [the localization plan](https://github.com/krambertech/lymi/blob/main/docs/plans/2026-09-12-localization.md): every learner-visible string in the product is a Lingui message (342 extracted), plurals, dates and language names follow the active locale, and `appLanguage` runs as a closed enum from core through schema, service, route and MCP with `meaningLanguage` derived from it and no longer writable. The client picks the language before first render, the bare shell follows the browser, sign-out clears the stored choice, and push reminders read `app_language`. Nothing visible changes in English and no learner data changes: the picker and the browser-pick write wait for phase 2.

## Why

Two things the spike found are now guarded rather than remembered. TanStack's split route modules end in `?tsr-split=component`, which the Babel filter missed, so their macros reached the bundle intact and would have thrown on first render in production; the filter is widened and the deployment-boundary check now fails on an uncompiled macro. The Lingui CLI needs Node 22.19, so `.nvmrc` and `engines.node` move up and `i18n:check` in `verify` fails on an unextracted string.

## Verification

- `pnpm verify`: pass (Biome, `i18n:check`, migrations, both builds, boundaries, typecheck, 114 tests).
- `pnpm deploy:check`: both dry runs pass.
- `pnpm test:e2e:chromium`: 14 passed. WebKit not run; no navigation, caching or sheet behavior changes.
- Built client assets contain no Lingui macro runtime (they did before the filter fix).

## Follow-ups

Phase 2 adds the Ukrainian and Russian catalogs, the **App language** picker, `failOnMissing`, and the language e2e. The `translate` skill and glossary are in this PR so that work can start from it.


apps/web/src/server/push-delivery.ts
```diff
@@ -68,21 +70,24 @@ export function reminderIsDue(
...
-export function reminderCopy(due: number) {
+export async function reminderCopy(due: number, locale = "en") {
+  const i18n = await serverI18n(locale);
   return {
-    title: due === 1 ? "One word is ready" : "A few words are ready",
-    body:
-      due === 1
-        ? "One card is waiting when you have a moment."
-        : `${due} cards are waiting when you have a moment.`,
+    title: due === 1 ? i18n._(msg`One word is ready`) : i18n._(msg`A few words are ready`),
```
krambertech, line 77: The title branches on due === 1, while Ukrainian uses the one category for values such as 21 and 31 too. reminderCopy(21, "uk") currently returns a singular body with the plural title “Кілька слів готові”. Could we express the title through plural as well and assert the complete copy for 1, 2, 5, and 21?

apps/site/src/pages/uk/index.astro
```diff
@@ -0,0 +1,27 @@
...
+  name: "Lymi",
+  applicationCategory: "EducationalApplication",
+  operatingSystem: "Web",
+  description: LANDING_BLURB,
+  url: "https://lymi.app/",
+  offers: { "@type": "Offer", price: "0", priceCurrency: "USD" },
+};
+---
+
+<BaseLayout title={LANDING_TITLE} description={LANDING_BLURB} canonicalPath="/uk/" dark>
```
krambertech, line 18: This page renders /uk/ as its canonical URL, but the hydrated LandingView effect overwrites every existing canonical link with the site root. Could we preserve the server-provided canonical or pass the localized path into the island, and cover the DOM after hydration?

PR #50, opened 2026-09-12
feat(web): add local personas, a developer panel and pnpm local

krambertech:
## ✨ What

Local testing no longer needs Google or a typed password. Opening `http://localhost:5241/api/dev/sign-in?as=learner` creates and seeds a persona account and lands on Today. A developer panel in the product (bottom-right button, or the backtick key) switches persona, sets how many cards are due, reseeds or empties the account, and changes meaning language and theme without a reload. `pnpm local` does the same from a terminal. Everything is gated on a loopback `PRODUCT_URL`, so production answers 404 to all of it and the panel is not in the production bundle.

## ✅ How to verify

- Start the product (`pnpm --filter @lymi/web dev --port 5241`), open `/api/dev/sign-in?as=streak`, and see Today with a 14-day streak and nothing due. Press backtick and pick another persona.
- `pnpm local personas`, then `pnpm local due 5 --as learner` and watch Today update.
- `pnpm verify` passes (95 unit tests in apps/web, 21 in core, 24 script tests). `pnpm deploy:check` passes. `pnpm test:e2e` passes in Chromium and WebKit (26 passed, 2 skipped).

## 🤔 Why

Sign-in was the step agents and people alike got stuck on, and a fresh account cannot show a streak, a backlog, cards an assistant added, or Ukrainian meanings without minutes of setup. Personas are separate accounts in the local D1 with a deterministic seed, so the same screen state is one URL away every time. Seeding goes through the real deck and card services, so the data carries audit rows and scheduling state like real data, and the review history is replayed in memory through the real scheduler and written back in one go.

The Combobox change is not dev-only. Its list used to unroll in flow so it could survive the phone drawer's transform and the desktop `<dialog>`; putting it in the browser's top layer through the Popover API lets it float in both places, so the new-deck sheet and deck settings get the same fix as the panel.

## 📸 Screenshots

The panel above its corner button, with the persona list floating over the rows. Desktop, light.

Desktop, light, persona list open

<details>
<summary>Desktop light and dark, and the phone</summary>

Desktop, light

Desktop, dark

Phone, light

</details>

## 📝 Notes

- Today counts due states while the review queue counts cards, so a deck asked both ways shows a higher number on Today than the session holds. Pre-existing; the panel's due control speaks in cards.
- The worktree's local D1 carried migrations from an older branch under different names. `pnpm local db:fresh` is the recovery: it moves the state directory aside and applies every migration again.


apps/web/src/server/index.ts
```diff
@@ -17,6 +17,7 @@ import { audio } from "./routes/audio";
 import { cards } from "./routes/cards";
 import { connectedApps } from "./routes/connected-apps";
 import { decks } from "./routes/decks";
+import { dev } from "./routes/dev";
```
krambertech: This static import pulls the dev routes, persona fixtures, fixed password, and reset/seed services into `dist/lymi/index.js`; the production artifact contains `/api/dev`, `lymi-dev-password`, and `resetAccount`. The runtime `PRODUCT_URL` gate returns 404 under the current production config, but this still violates the stated boundary that the local tooling never reaches production and leaves destructive code in the deployed Worker. Use a separate local Worker entry/config or a build-time conditional so production cannot import these modules, keeping the runtime gate as defense in depth.

apps/web/src/client/dev/session-guard.ts
```diff
@@ -0,0 +1,27 @@
...
+ * right for one learner and wrong the moment the developer becomes another one: Today would
+ * hydrate the previous persona's decks and name until each query refetched. The dev sign-in
+ * sets a `lymi_dev_persona` cookie; when it differs from the one last seen here, the cache
+ * and the offline outbox are thrown away before the app mounts.
+ */
+const SEEN_KEY = "lymi-dev-persona";
+const PERSISTED = ["lymi-query-cache", "lymi-outbox"];
+
+export function guardPersistedCache(): void {
+  const current = /(?:^|;\s*)lymi_dev_persona=([^;]*)/.exec(document.cookie)?.[1] ?? "";
```
krambertech, line 15: Cache eviction is keyed only to `lymi_dev_persona`, but normal sign-out and Google or dev-email sign-in never clear that cookie. After signing out of a persona and into a real local account, the cookie and `SEEN_KEY` still match, so the previous account cache and offline outbox hydrate into the new session; `me` can remain fresh for five minutes. Key the guard to every authenticated user/session, or clear the marker and persisted data on every non-persona sign-in and sign-out path.

apps/web/src/server/services/dev.ts
```diff
@@ -0,0 +1,332 @@
...
+      id: schema.cardStates.id,
+      cardId: schema.cardStates.cardId,
+      due: schema.cardStates.due,
+      lastReview: schema.cardStates.lastReview,
+    })
+    .from(schema.cardStates)
+    .innerJoin(schema.cards, eq(schema.cards.id, schema.cardStates.cardId))
+    .innerJoin(schema.decks, eq(schema.decks.id, schema.cards.deckId))
+    .where(
+      and(
```
krambertech, line 256: This query includes every state on active cards but omits the shared `asked` predicate. Lymi deliberately retains inactive direction states when a deck changes direction, so those rows can consume the requested card count: `due 5` may return 5 while producing fewer than five reviewable cards. Filter to asked states before selecting or changing due dates.

apps/web/src/server/services/dev.ts
```diff
@@ -0,0 +1,332 @@
...
+    }
+  }
+
+  const statements: unknown[] = [];
+  if (dueIds.length > 0) {
+    statements.push(
+      db
+        .update(schema.cardStates)
+        .set({ due: new Date(now.getTime() - 60_000), updatedAt: now })
+        .where(inArray(schema.cardStates.id, dueIds)),
```
krambertech: `DueBody` accepts up to 10,000 cards and the CLI offers `all`, but this update binds every due state ID in one `IN` clause. The same file documents D1’s 100-parameter limit, so an account over roughly 100 active states makes `due all` or a large count fail. Chunk these IDs into bounded updates before passing them to `runBatched`.

apps/web/src/client/dev/DevPanel.tsx
```diff
@@ -0,0 +1,352 @@
...
+        target &&
+        (target.tagName === "INPUT" ||
+          target.tagName === "TEXTAREA" ||
+          target.tagName === "SELECT" ||
+          target.isContentEditable);
+      if (e.key === "`" && !typing && !e.metaKey && !e.ctrlKey && !e.altKey) {
+        e.preventDefault();
+        setOpen((v) => !v);
+      }
+      if (e.key === "Escape" && open && panelRef.current?.contains(document.activeElement)) {
```
krambertech, line 89: Escape or backtick closes the panel while focus is inside it, but unmounting the focused section does not return focus to the toggle. A keyboard user is left on the document body, so the next Tab starts from the page rather than the control they used. Keep a trigger ref and restore focus after keyboard dismissal.

krambertech:
Review verdict: changes needed.

I left five inline findings:
- the production Worker artifact still contains the developer routes, personas, fixed password, and reset tooling;
- cache isolation fails when moving from a persona to a real local account;
- due-count selection includes inactive direction states;
- large or `all` due updates can exceed D1’s bind-parameter limit;
- keyboard dismissal of the developer panel does not restore focus.

Verification:
- GitHub CI is green at `6978592f`, including the base gate, both deployment-package checks, and 14 Chromium E2E tests.
- Locally, `pnpm verify` and `pnpm deploy:check` passed; the focused gate/persona tests (6) and CLI helper tests (4) passed.
- The production client bundle contains no developer panel or persona cache-guard code. The production Worker bundle does contain `/api/dev`, the fixed developer password, and reset services.
- With production URL configuration, `/api/dev/personas` returned 404. With loopback configuration, the read-only personas endpoint returned 200 with the expected five personas.

Not independently re-verified: mutating seed/reset/due flows, rendered mobile/desktop/theme/accessibility states, and WebKit. A full local browser run was invalidated by another concurrent E2E server owning the same ports, so I did not treat those failures as PR regressions. The PR’s CI Chromium run remains green; WebKit is supported only by the author’s stated local result.

apps/web/src/server/index.ts
krambertech: Fixed in 9f625b6. The import now sits inside an `import.meta.env.DEV` branch, so the production Worker bundle contains no `/api/dev` route, no persona fixture, no `lymi-dev-password`, and no seed or reset service (checked by grepping `dist/lymi/index.js` after `pnpm build`). What remains in the bundle is the runtime gate itself: the loopback check and the `@lymi.local` allowlist exception, kept as defence in depth. `docs/local-dev.md` describes both gates.

apps/web/src/client/dev/session-guard.ts
krambertech: Fixed in 9f625b6. New `lib/persisted.ts` clears the query cache, the offline outbox and the persona marker in one call. Sign-out on the You screen calls it after `queryClient.clear()`, Google sign-in calls it before redirecting, and the local email sign-in calls it on success. Verified in the browser: after signing out of a persona the persisted cache holds zero queries and the marker is gone.

apps/web/src/server/services/dev.ts
krambertech: Fixed in 9f625b6. The state query now carries the shared `asked` predicate from the decks service, so states for a direction the deck no longer asks in are neither counted nor moved.

apps/web/src/server/services/dev.ts
krambertech: Fixed in 9f625b6. The due ids go into `IN` updates in slices of ninety, each its own statement, before `runBatched`. `pnpm local due all --as backlog` and `due 3 --as learner` both run against the local D1.

apps/web/src/client/dev/DevPanel.tsx
krambertech: Fixed in 9f625b6. The trigger button has a ref; Escape and the backtick key both go through a `dismiss` that closes the panel and returns focus to the button on the next frame. Verified: after Escape, `document.activeElement` is the Developer tools button.

krambertech:
Addressed all five findings in 9f625b6. Each thread has the specific change.

Verification on the new head: `pnpm verify`, `pnpm deploy:check`, and `pnpm test:e2e` in Chromium and WebKit (26 passed, 2 skipped). The production Worker bundle was grepped after the build: no `/api/dev`, no fixed password, no seed or reset code; only the loopback gate remains.

The port clash you hit locally was a stale E2E server from an earlier run holding 4173 to 4175, not a PR regression. I hit the same thing twice; freeing those ports before `pnpm test:e2e` clears it.

krambertech:
Rebased onto `main` after the Lingui and Library changes landed; the conflict is cleared. Head is 9102c6f.

One follow-on: `main` now derives the meaning language from a new interface language setting, so a persona declares `appLanguage` and the panel's language row is a segmented control over English, Ukrainian and Russian. The `polyglot` persona switches the app to Ukrainian; its sentences stay English until the catalog is filled, which is outside this PR.

On the rebased head: `pnpm verify`, `pnpm deploy:check`, and `pnpm test:e2e` in Chromium and WebKit (28 passed, 2 skipped). The production Worker bundle still carries none of the developer routes or fixtures.

PR #51, opened 2026-09-12
feat(decks): add deck membership and per-learner states

krambertech:
## Summary

Step 1 of [the shared decks plan](docs/plans/2026-09-12-shared-decks.md): the data model and read paths for a deck with members, behind no interface yet. A `deck_members` table records who studies a deck besides its owner, `card_states` becomes unique per learner, every read goes through membership, and every write to a deck's content requires the owner. Adding a card or turning on a direction creates states for every learner. Deck responses now carry `role` and `owner`. Decided in [ADR 0011](docs/adr/0011-a-shared-deck-is-one-deck-with-many-learners.md).

## Verification

- `pnpm verify` passes, including ten new service tests in `members.test.ts` on a real local D1: a member sees the deck, cards, search, and their own due count; grades without touching the owner's schedule; gets 403 on every write; receives new cards and direction changes; a thirty-card lesson lands whole for three learners across batches; rejoin works after leaving and not after removal.
- `pnpm deploy:check` passes.
- `pnpm test:e2e:chromium` passed on the pre-rebase head, 14 journeys. After rebasing onto #42 the local run could not start because another session holds port 4173, so CI's Chromium job is the browser evidence for this head. WebKit was not run; nothing here touches navigation, PWA, touch, or dialogs.

## Why

Three things fell out of building it. Card adds now batch in whole per-card groups of about fifty statements, since one D1 batch cannot hold a shared lesson and a card must never land without its states. The state fill chunks learners and cards so a large class stays under D1's hundred bound parameters. The audio key is a cache on the owner's card, so its write and audit row use the card's owner rather than whoever pressed play.

## Risks

The migration swaps the `card_states` unique index and moves no rows. Existing rows already carry the right `user_id`.

## Follow-ups

- PR 2 adds invitations, the sign-in gate, and the join page. `join`, `leave`, `removeMember`, and `listMembers` exist as services but have no routes yet.
- `cardHistory` from #42 lists audit events by the caller's user id, so a member sees their own reviews on a shared card but none of the owner's edits. Whether members should see those is a PR 3 decision.


apps/web/src/server/services/cards.ts
```diff
@@ -38,25 +40,25 @@ export async function addCards(
...
-          inArray(schema.decks.id, ids),
-          eq(schema.decks.userId, userId),
-          isNull(schema.decks.archivedAt),
-        ),
-      ),
+      .where(and(inArray(schema.decks.id, ids), memberOf(userId), isNull(schema.decks.archivedAt))),
   );
   const deckById = new Map(decks.map((d) => [d.id, d]));
+  const learnersByDeck = new Map<string, string[]>();
+  for (const deck of decks) learnersByDeck.set(deck.id, await learnersOf({ db }, deck.id));
```
krambertech: This learner snapshot races with join: if add reads members before join inserts, while join reads cards before this batch inserts, both operations succeed but the new member never gets a state for the card. Fan out from live membership in the same D1 transaction as the card insert, with the symmetric insert-from-current-cards on join, so either serialization order creates the state.

apps/web/src/server/services/cards.ts
```diff
@@ -307,6 +344,10 @@ export async function updateCard(ctx: ServiceContext, id: string, patch: CardPat
       updatedAt: new Date(),
     })
     .where(eq(schema.cards.id, id));
+  const askedChanged =
+    (patch.directions !== undefined && patch.directions !== current.directions) ||
+    (patch.deckId !== undefined && patch.deckId !== current.deckId);
+  if (askedChanged) await openCardDirections(ctx, id);
```
krambertech: The card mutation commits before fan-out, and a retry with the same patch makes askedChanged false. A partial fillStates failure can therefore persist the new deck or direction while permanently skipping some learners; run the idempotent backfill whenever either field is present, or make mutation and fan-out atomic and recoverable.

apps/web/src/server/services/members.ts
```diff
@@ -0,0 +1,295 @@
...
+async function askedInDeck({ db }: Pick<ServiceContext, "db">, deckId: string) {
+  const [deck] = await db
+    .select({ directions: schema.decks.directions })
+    .from(schema.decks)
+    .where(eq(schema.decks.id, deckId));
+  if (!deck) throw notFound("Deck");
+  const cards = await db
+    .select({ id: schema.cards.id, directions: schema.cards.directions })
+    .from(schema.cards)
+    .where(and(eq(schema.cards.deckId, deckId), isNull(schema.cards.archivedAt)));
```
krambertech: Excluding archived cards here leaves a permanent state gap: a learner who joins while a card is archived gets no state, and restoreCard only clears archivedAt, so the restored card never enters their queue. Either create states for archived cards during join or backfill every active learner on restore.

apps/web/src/server/services/members.ts
```diff
@@ -0,0 +1,295 @@
...
+    .from(schema.decks)
+    .where(and(eq(schema.decks.id, deckId), isNull(schema.decks.archivedAt)));
+  if (!deck) throw notFound("Deck");
+  if (deck.userId === userId) return { ok: true as const, role: "owner" as const };
+
+  const [existing] = await db
+    .select()
+    .from(schema.deckMembers)
+    .where(and(eq(schema.deckMembers.deckId, deckId), eq(schema.deckMembers.userId, userId)));
+  if (existing && !existing.removedAt) return { ok: true as const, role: existing.role };
```
krambertech: An active row is treated as a completed join, but it is persisted before state fan-out and the audit write. If either later step fails, a retry returns here and permanently leaves an active member with missing states or no join event; reconcile missing states before returning, or make the logical join atomic and recoverable.

krambertech:
Reviewed exact head `57b149b3` against ADR 0011 and the shared-decks plan. I posted four inline state-consistency findings: join retry recovery, archived-card restore, the add/join race, and card-update retry recovery.

Verification:
- `pnpm verify` passed locally (143 tests across package, app, and script suites).
- `pnpm deploy:check` passed for both packages; no production action was taken.
- Current-head CI passed, including Chromium E2E and both deployment-package checks. My independent Chromium run was inconclusive because another workspace repeatedly occupied the fixed harness ports: 6 tests passed before the first collision; the clean retry was blocked before startup.
- Migration history, schema drift, mergeability, and existing review threads were checked. The PR is cleanly mergeable and had no prior human reviews or inline comments.

The auth/privacy trace confirmed owner-only content writes and caller-scoped scheduling/history. Invitation validation remains a required boundary for the later external join surface; member-roster visibility and production D1 query-budget behavior should also be settled or exercised before those surfaces ship.

PR #53, opened 2026-09-12
feat(web): fluid hover fill and popover Select and Combobox

krambertech:
## Summary

- Menus and option lists get one hover fill that slides to the nearest enabled row instead of blinking off between rows. It skips disabled rows, never crosses a separator, hides the moment the keyboard takes over, and only fades under reduced motion.
- `Select` and `Combobox` are now one picker. The box stays a button; the list floats under it as a top-layer popover (so the form does not move and it escapes vaul's transformed drawer), search sits inside the panel, the panel flips above the box when there is more room there, and the check leads each row in a slot every row keeps. The native `<select>` is gone.
- A pass over the rest of the component set: toasts pause while hovered or focused so Undo cannot leave under a hand reaching for it, Checkbox and Switch rows reach 44 px on the phone, menus take Home and End, Table aligns logically like the rest of the client, and the Copy button keeps one width as its label changes. The design page gains a Select and Combobox section and its frames no longer overflow a phone.

## Verification

- `pnpm verify`: pass.
- `pnpm test:e2e:chromium`: pass.
- Checked in a scripted browser: zero layout shift when a picker opens, panel on top of the desktop `<dialog>` sheet and the phone drawer, search focused on open, a press on the panel's scrollbar keeps it open, focus returns to the box, outside click and Escape close, keyboard walking and type-ahead pick the right row, Home and End in menus, 44 px Checkbox and Switch rows at 390 px, design page frames within the phone viewport.
- WebKit and mobile projects were not run. The toast pause is code-verified only.

## Risks

Any test or integration that read the deck picker as a native `<select>` value now reads its visible text; `e2e/deck-creation.spec.ts` is updated accordingly.


apps/web/src/client/components/Combobox.tsx
```diff
@@ -11,14 +23,17 @@ export interface ComboboxOption {
   hint?: string | undefined;
 }
 
-interface Props {
+interface BaseProps {
```
krambertech, line 26: Every new interactive control is required by this PRs design text to support a disabled state, but this API has no disabled prop and the trigger always remains operable. Replacing the native Field.Select therefore removes its disabled semantics entirely; add a disabled picker prop, pass it to the trigger, and prevent opening or selection while disabled.

apps/web/src/client/components/Combobox.tsx
```diff
@@ -189,87 +318,106 @@ export function Combobox({
...
+          <div className="flex shrink-0 items-center gap-2.5 border-b border-edge ps-3.5 pe-3.5">
+            <Search className="size-4 shrink-0 text-muted" aria-hidden="true" />
+            <input
+              ref={searchRef}
+              role="combobox"
+              aria-expanded={open}
+              aria-controls={listId}
+              aria-autocomplete="list"
+              aria-activedescendant={activeId}
+              aria-label={searchLabel}
```
krambertech, line 371: When this Combobox opens, focus moves to this input, but the Field metadata is left on the trigger. I reproduced the error specimen in Chromium and WebKit: the trigger has aria-invalid=true and its error aria-describedby, while the focused search input has neither, so the active control no longer exposes the error or hint; propagate those attributes to the search input while it owns focus.

krambertech:
Review complete. I left two actionable inline comments: the custom picker contract drops the required disabled state, and the focused Combobox search input loses its Field error and hint semantics.

Verification: pnpm verify, pnpm deploy:check, focused fluid-hover unit tests (7 of 7), and pnpm test:e2e (28 passed, 2 intentionally skipped) all passed. I also exercised Select, Combobox, and Menu in Chromium and WebKit across pointer and keyboard input, filtering and empty results, selection, dismissal and focus return, light and dark specimens, small-viewport flipping, and reduced motion; no runtime errors appeared.

The signed-in mobile drawer harness did not open the Add menu reliably, so that specific picker-in-drawer interaction remains unconfirmed. The iPhone WebKit E2E suite passed, but it does not directly select a deck from that drawer.

PR #56, opened 2026-09-12
feat(web): replace You with Settings and a learner menu

krambertech:
## Summary

- `/you` becomes `/settings` and holds only settings: language, theme, the daily reminder, connected apps and API keys. `/you` redirects.
- A learner menu behind the name in the rail and the avatar on the phone carries Settings, Archived, Docs and Sign out; the phone's menu also carries Activity and Insights, and the rail's carries Keyboard shortcuts. Install Lymi appears only where the browser can do it.
- Connected apps and API keys get empty states that point at the docs, key creation opens as a sheet, and the reminder row shows load problems as state instead of a red alert.

## Verification

- `pnpm verify`: pass
- `pnpm test:e2e:chromium`: 15 passed
- Manual: rail and phone menus, keyboard shortcuts dialog, sign-out flow, language switch, key creation and revocation, light and dark themes

## Why

The rail shows the first name only, and sign-out is red because it drops queued offline grades and needs Google to get back in; the outbox is flushed before signing out. Language switching works, but the Ukrainian and Russian catalogs are still almost empty, so the interface stays English after switching while meanings follow the choice.

## Screenshots

Not attached yet; handed over in the session.


e2e/core-learning-flow.spec.ts
```diff
@@ -12,10 +12,11 @@ test("a protected deep link survives sign-in", async ({ page }, testInfo) => {
   await expect(page).toHaveURL(/\/library\?from=reminder&deck=italian$/);
   await expect(page.getByRole("heading", { name: "Library", exact: true })).toBeVisible();
 
-  await page.goto("/you");
-  await expect(page.getByRole("link", { name: "Lymi website" })).toHaveAttribute(
+  // The docs link is the one place the product points at the public site's configured origin.
+  await page.getByRole("button", { name: "Dev", exact: true }).click();
```
krambertech, line 18: [P1] Open the phone menu from a route that actually has it

The WebKit project uses the iPhone 15 profile, and `/library` has neither the desktop rail nor the Today avatar, so this locator never exists; `pnpm test:e2e` fails here deterministically. Navigate to `/today` before opening the menu (or choose the entry point per project) so the required mobile gate passes and the phone learner menu is genuinely covered.

apps/web/src/client/lib/use-sign-out.ts
```diff
@@ -0,0 +1,31 @@
...
+  const queryClient = useQueryClient();
+  const [busy, setBusy] = useState(false);
+  return {
+    busy,
+    signOut: async () => {
+      if (busy) return;
+      setBusy(true);
+      try {
+        // Grades made offline would leave with the session; send them first if the network is back.
+        await flushOutbox().catch(() => 0);
```
krambertech: [P0] Preserve queued grades when sign-out cannot flush them

`flushOutbox` keeps network-failed grades in `lymi-outbox` and still resolves, so this continues into sign-out and `clearPersistedLearnerState()` deletes them. Going offline after grading and choosing Sign out therefore permanently loses those reviews; keep the learner signed in until the outbox is empty, or require an explicit data-loss choice that preserves the default safe path.

apps/web/src/client/views/SettingsView.tsx
```diff
@@ -0,0 +1,90 @@
...
+      />
+
+      <SettingsGroup
+        title={t`Language`}
+        description={t`For the interface, and for the meanings the AI writes.`}
+      >
+        {language ? (
+          <div className="w-56">
+            <Select
+              label={t`Language`}
```
krambertech: [P1] Give the language combobox an accessible name

This `label` names the popup listbox, not the closed combobox trigger; rendered inspection exposes the trigger only as its current value, and `getByRole("combobox", { name: "Language" })` finds nothing. Wrap the picker in `Field` so its generated label targets the trigger, or add an equivalent accessible-name connection.

krambertech:
Cross-cutting documentation gap: `CONTEXT.md` now makes Settings and the learner menu canonical, but `DESIGN.md`, ADR 0005 and its index, `docs/site-structure.md`, `docs/testing.md`, and ADR 0013 still describe the old You screen or omit `/settings` from the product-origin contract. Please update the owning current-behavior documents and preserve or supersede the historical ADR as appropriate, so the repository has one navigation and route contract.

e2e/core-learning-flow.spec.ts
krambertech: Fixed in e557f25: the test goes to /today before opening the menu, so the rail on desktop and the avatar on the phone both carry it.

apps/web/src/client/lib/use-sign-out.ts
krambertech: Fixed in e557f25: sign-out flushes the outbox and stops if anything is still queued, with a dialog whose default is Stay signed in; losing the grades is an explicit second button. Checked locally by seeding two queued grades with the grade endpoint failing.

apps/web/src/client/views/SettingsView.tsx
krambertech: Fixed in e557f25: the picker sits in a Field whose label is visually hidden, so the trigger is named Language and the language e2e now locates it by name.

PR #63, opened 2026-09-13
feat(web): redesign the deck page around a review plate

krambertech:
## Summary

- A deck now opens on a plate: cards due now set large, the deck's New / Learning / Known counts with icons, and Review (or Add card when nothing is due, with when the next card is back). The stripe under it has no heading or legend; the filter below names the states with dots. Rename leaves the deck menu, since deck settings already has it.
- Card states get their own tokens, used everywhere a state shows: grey New, yellow Learning, green Known, each at 3:1 or better against plate and canvas. New chips were amber, which DESIGN.md already ruled out for state.
- Phone screens share one `TopBar` (Today, Library, deck, deck settings, card page): back as a 44 px target naming its destination, square icon buttons, round amber capture first. Library's bar gains the avatar to match Today.

## Why

A deck's `due` now counts cards rather than direction states, in the API and MCP too. A session asks one direction per card, so a deck studied both ways used to show twice the cards Review would start, on Today and the rail as well as the plate.

## Verification

- `pnpm verify`: passed
- `pnpm deploy:check`: passed
- `pnpm test:e2e` (Chromium and WebKit): 32 passed, 2 skipped
- Checked by hand on phone and desktop widths, light and dark, with the `learner` persona

## Follow-ups

- Insights, Settings, Activity and Archived don't use `TopBar` yet, so their titles sit higher on the phone.


apps/web/src/client/views/DeckDetailView.tsx
```diff
@@ -114,31 +112,128 @@ function reps(state: CardState | null): number {
...
+}: {
+  deck: DeckSummary;
+  cards: Row[];
+  /** The whole deck by state: new, learning, known. */
+  counts: Record<0 | 1 | 2, number>;
+  onReview?: (() => void) | undefined;
+  onAdd: () => void;
+}) {
+  const { t, i18n } = useLingui();
+  const due = deck.due;
```
krambertech, line 152: [P1] `deck.due` counts due direction-state rows, while the plate labels it as cards and the state totals contain one row per card. A two-card deck using both directions can therefore show “4 cards due” beside totals adding up to 2, then open a queue containing one direction per card; derive all plate figures from a consistent per-card summary.

apps/web/src/client/styles.css
```diff
@@ -43,6 +43,13 @@
 
   --good: oklch(0.47 0.12 160);
   --good-soft: oklch(0.47 0.12 160 / 0.12);
+  /* Card states, the same three everywhere: grey not yet asked, yellow on the way, green known.
+     Learning sits well off amber's hue so a state never reads as something to press. */
+  --state-new: oklch(0.74 0.012 60);
```
krambertech: [P1] Against the white plate, `state-new` is about 2.31:1 and `state-learning` about 1.68:1, below the 3:1 non-text contrast floor. These colors communicate state in stripes, dots, and icons, so darken them or add a sufficiently contrasting visual boundary.

apps/web/src/client/views/DeckDetailView.tsx
```diff
@@ -114,31 +112,128 @@ function reps(state: CardState | null): number {
...
+  ];
+
+  return (
+    <section className="edge grid rounded-2xl bg-plate px-5 pt-7 pb-5 text-center @3xl:grid-cols-[auto_minmax(0,1fr)_auto] @3xl:items-center @3xl:gap-8 @3xl:px-8 @3xl:py-7 @3xl:text-start">
+      <h2 className="grid justify-items-center gap-1.5 @3xl:justify-items-start">
+        <span className="text-5xl font-semibold tracking-[-0.03em] tabular-nums">
+          {i18n.number(due)}
+        </span>
+        <span className="text-md text-text-2">
+          <Plural value={due} one="card due today" other="cards due today" />
```
krambertech: [P2] The server counts states whose due time is at or before the current instant, not every card due during today. A card scheduled for later today makes this plate say “0 cards due today” immediately above “The next card is back in 4 hours”; use “due now” or change the underlying time window consistently.

PR #64, opened 2026-09-13
feat(review): add motion to the review screen

krambertech:
## Summary

- Reveal is now one choreographed moment built with Motion: the word glides up, the rule draws, the answer staggers in, and the grade strip opens as the card shrinks. Keyboard reveals stay instant.
- The top bar puts the lantern, track, rolling count and a quiet close button on one line. The state chip is larger, and a lapsed card says "Forgot recently".
- A tap hint shows only on a learner's first three cards and after a minute idle. A failed pronunciation turns its button red, shakes it, and explains in a tip instead of adding a line under the card.

Also fixes "Add a word" on Today and Library passing the click event as a deck id, which crashed the capture sheet.

## Verification

- `pnpm verify`: passed (web unit tests 125, including the new hint-delay tests)
- `pnpm test:e2e:chromium`: 17 passed; the full cross-browser `pnpm test:e2e` was not run
- Browser checks at 375 px and 1440 px: reveal and card-shrink frames sampled, hint timing for new and idle learners, forced audio failure with tip dismissal, dark mode

## Risks

The reveal hint's reveal count lives in `localStorage` and clears on sign-out, so a learner sees the first-run hint again on a new device.


apps/web/src/client/routes/review.tsx
```diff
@@ -173,10 +181,11 @@ function Review() {
   }, [revealed, onGrade, navigate]);
 
   return (
-    <div className="mx-auto flex min-h-0 w-full max-w-md flex-1 flex-col px-4 pb-safe @3xl:max-w-xl @3xl:px-8 @3xl:pb-8 @3xl:pt-4">
+    <div className="mx-auto flex min-h-0 w-full max-w-md flex-1 flex-col px-4 pb-[calc(env(safe-area-inset-bottom)+12px)] @3xl:max-w-2xl @3xl:px-8 @3xl:pb-8 @3xl:pt-4">
```
krambertech, line 184: This widens the desktop review from 576 px to 672 px, while the card below is raised from the documented 440 px cap to 460–600 px. The live Layout specimen still defines a 576 px review column and a card no taller than 440 px, so the implementation and owning design system now disagree. Please either keep the existing geometry or update the layout contract explicitly if this resize is intended.

apps/web/src/client/styles.css
```diff
@@ -476,6 +476,43 @@
   }
 }
 
+/* ---- Error tip --------------------------------------------------------- */
+
+/* Rises off the control it describes and leaves faster than it came, like the picker panel. */
+.error-tip {
```
krambertech, line 503: The tip scales from the browser default centre, so it does not visually rise from the control it points to. The below-anchor fallback also reverses only the exit: the starting style still begins at translateY(4px), moving in from away from the button. Please use bottom/top-centre origins and mirror the entry direction when data-below is set.

PR #65, opened 2026-09-13
feat(web): make the daily goal the streak requirement

krambertech:
## Summary

- Step 1 of [the daily review goal plan](docs/plans/2026-09-13-daily-review-goal-and-rolling-queue.md): a learner-local review day counts toward the streak when its goal's attempts land (every accepted grade, Forgot and repeats included) or when every eligible review is done below it. Opening Lymi on a day with nothing due confirms it, which keeps the run without adding to it. Reviews from before goals still count as reviewed days.
- New `review_days`, `review_undos`, and `reviews.review_day_id` / `state_before`; `POST /api/review/undo` restores the card and takes the attempt back inside one transaction that re-checks the grade is still latest; `POST /api/review/today` settles today; the review timezone follows the visible device unless set manually.
- The streak pill and modal read goal days; each past light is graded against its own day's goal and labelled with learner-local dates. The goal (10/25/50/100 or 1–200, default 50) is changed only in the modal, only by the learner, and its writes are serialized so the last choice is the last write. Today and review completion show the run with the seven lights.

## Screenshots

| Today, phone | Streak modal, phone (dark) | Streak modal, desktop |
| --- | --- | --- |
| Today on a phone | Streak modal on a phone | Streak modal on desktop |

## Verification

- `pnpm verify`: passed (web 144 tests, including service tests for goal, exhausted, nothing-due, missed day, undo, undo racing a newer grade, duplicate, timezone and pre-goal history).
- `pnpm deploy:check`: passed.
- E2E, Chromium and WebKit: 32 passed, 2 skipped.

## Risks

- Migration `0009` adds two tables and columns, so `db:migrate:prod` is needed on deploy.
- The first-review goal prompt, server-owned session, Undo in the review UI, and the Settings timezone control are later steps of the plan.


apps/web/src/server/mcp/server.ts
```diff
@@ -333,7 +336,7 @@ export function buildMcpServer(principal: McpPrincipal): McpServer {
     {
       title: "Change settings",
       description:
-        "Change the app language, which also sets the language meanings are written in. Cards already written are not translated. Needs write.",
+        "Change the app language, which also sets the language meanings are written in, or the daily goal: how many cards make a full day. Cards already written are not translated. Needs write.",
```
krambertech: [P1] Disclose settings changes in the write grant

This extends `update_settings` so an MCP client can change the learner’s daily goal, but the consent screen still describes write access only as “Add, edit and archive cards” and says everything the app can do is listed there. Either keep settings learner-only or update the consent/scope contract before exposing this write, so granting card access does not silently authorize goal changes.

apps/web/src/client/components/StreakCalendar.tsx
```diff
@@ -0,0 +1,198 @@
...
+                      <i
+                        aria-hidden="true"
+                        className="absolute top-1/2 left-1/2 block size-8 -translate-1/2 rounded-full edge-2"
+                      />
+                    )}
+                    <span
+                      aria-hidden="true"
+                      className={clsx(
+                        "relative text-sm tabular-nums",
+                        on ? "font-medium text-text" : future ? "text-faint" : "text-muted",
```
krambertech: [P2] Keep future date text above the contrast floor

`text-faint` is explicitly decorative-only in DESIGN.md and does not meet the 4.5:1 text contrast floor, but these future day numbers are meaningful calendar text for sighted learners. Please use a passing text token, such as `text-muted`, and convey “future” without dropping the dates below the readability floor.

apps/web/src/client/lib/streak.tsx
```diff
@@ -0,0 +1,56 @@
...
+      if (settings) qc.setQueryData<Settings>(settingsQuery.queryKey, { ...settings, dailyGoal });
+      return { streak, settings };
+    },
+    onError: (_error, _goal, context) => {
+      if (context?.streak) qc.setQueryData(streakQuery.queryKey, context.streak);
+      if (context?.settings) qc.setQueryData(settingsQuery.queryKey, context.settings);
+      setStatus("error");
+    },
+    onSuccess: (settings) => {
+      qc.setQueryData(settingsQuery.queryKey, settings);
```
krambertech, line 61: [P2] Make the latest goal choice win

GoalPicker stays enabled while this mutation is pending, so a custom blur followed by a preset tap can send two requests. An older response can then overwrite the newer settings cache, while the streak cache keeps the second optimistic value; an older error can also roll both caches back. Serialize goal changes or ignore stale mutation callbacks, then reconcile both query keys from the winning server response.

apps/web/src/client/components/Streak.tsx
```diff
@@ -0,0 +1,379 @@
...
+      </button>
+      {/* In the body, because a pill inside a hidden rail or header would hide an open modal with it
+          when the window crosses the breakpoint, and leave the page inert behind nothing. */}
+      {createPortal(
+        // biome-ignore lint/a11y/useKeyWithClickEvents: the native dialog closes on Escape; this click only catches the backdrop
+        <dialog
+          ref={dialogRef}
+          aria-labelledby={titleId}
+          onClick={(e) => {
+            if (e.target === dialogRef.current) close();
```
krambertech: [P2] Do not treat the phone sheet surface as a backdrop

On max-md the dialog itself fills the viewport, so this condition also closes the streak flow when the learner taps any unused space or padding inside the visible sheet. I reproduced it at 320×568 in the goal view by tapping below the options; limit backdrop dismissal to the centred desktop modal, or wrap the phone content so only a real backdrop can close it.

apps/web/src/client/components/Streak.tsx
```diff
@@ -0,0 +1,379 @@
...
+      />
+    );
+  }
+
+  const label = t`Streak: ${plural(summary.current, { one: "# day in a row", other: "# days in a row" })}`;
+  const face = clsx(
+    "relative inline-flex shrink-0 items-center rounded-full font-semibold text-text transition-[background-color,box-shadow,scale] duration-150 active:scale-[0.97]",
+    rail
+      ? "h-8 gap-1.5 px-2.5 text-sm before:absolute before:-inset-1.5 before:content-[''] hoverable:hover:bg-hover"
+      : "h-10 gap-2 bg-plate px-3.5 text-base edge hoverable:hover:bg-hover",
```
krambertech: [P2] Preserve the 44 px phone hit target

The phone variant is only 40 px high and has no expanded pseudo-element hit area, so this persistent control falls below Lymi’s documented 44 px phone target. Please make the button itself at least 44 px high or add an equivalent hit area.

apps/web/src/client/lib/streak.tsx
krambertech: The stale-callback guard fixes the local cache part, but it still does not order the server writes. If 25 and then 50 are sent and the 25 request finishes last, the server persists 25 while that callback is ignored and the UI remains at “Saved” for 50. Please serialize or coalesce these requests so the latest intent is also the final server write.

apps/web/src/server/services/review-days.ts
```diff
@@ -0,0 +1,386 @@
...
+      ctx.db
+        .update(schema.cardStates)
+        .set({
+          fsrs: before.fsrs,
+          due: new Date(before.due),
+          state: before.state,
+          lastReview: before.lastReview === null ? null : new Date(before.lastReview),
+          updatedAt: now,
+        })
+        .where(eq(schema.cardStates.id, review.state.id)),
```
krambertech: [P1] Make Undo conditional on the grade still being latest

The latest-grade check happens before this batch, while this update matches only the card-state ID. A newer grade can land between the check and the update, after which Undo restores the older snapshot and erases the newer schedule. Please make the rollback an atomic comparison against `lastReview` and insert the undo record only when that comparison succeeds.

apps/web/src/client/components/Streak.tsx
```diff
@@ -0,0 +1,444 @@
...
+  className?: string | undefined;
+}) {
+  const week = lastDays(summary);
+  const large = size === "lg";
+  return (
+    <div className={clsx("flex items-center", large ? "gap-5" : "gap-4", className)}>
+      <SevenLights
+        days={week.attempts}
+        satisfied={week.satisfied}
+        goal={summary.today.goal}
```
krambertech: [P1] Use each day’s snapshotted goal for its light

Every historical light is graded against today’s goal here. For example, a past 20/50 missed day becomes a full amber light after the learner lowers today’s goal to 10, so the history no longer reflects the outcome stored for that day. Please expose and use each day’s goal, or keep unsatisfied historical intensity independent of the current goal.

apps/web/src/client/components/SevenLights.tsx
```diff
@@ -56,7 +64,7 @@ const HEIGHT: Record<1 | 2 | 3, string> = { 1: "h-1/3", 2: "h-2/3", 3: "h-full"
  * One image, one description. The wrapper is atomic to assistive technology, so the per-day
  * counts go into its label rather than into titles nobody can reach from a keyboard.
  */
-export function SevenLights({ days, size = "sm", className }: Props) {
+export function SevenLights({ days, size = "sm", goal, satisfied, className }: Props) {
   const { t, i18n } = useLingui();
   const today = new Date();
```
krambertech, line 77: [P2] Label the actual learner-local dates

The counts end at `summary.today.date` in the learner’s review timezone, but these labels end at the device-local date. With a manual review timezone across a date boundary, both the visible weekdays and the accessible description are shifted by one day; please pass the actual dates selected by `lastDays` and format those.

apps/web/src/server/mcp/server.ts
```diff
@@ -320,7 +322,8 @@ export function buildMcpServer(principal: McpPrincipal): McpServer {
     "get_settings",
     {
       title: "Get settings",
-      description: "The learner's settings: the language meanings are written in.",
+      description:
+        "The learner's settings: the language meanings are written in, and the daily goal in cards.",
```
krambertech: [P2] Describe the goal in recall attempts

The accepted product contract counts every accepted grade, including Forgot and repeated cards, so ‘in cards’ teaches connected assistants the wrong unit. Please call this a daily goal in recall attempts.

krambertech (commented):
The earlier consent, contrast, modal, and hit-target findings are fixed at this head. The goal mutation race is only partially fixed, so I added a clarification to that thread. These inline comments cover the remaining backend, history, timezone, and terminology issues.

PR #73, opened 2026-09-13
feat(web): split the design system into routed pages

krambertech:
## Summary

- `/design` becomes a rail with one page per foundation, component group, and screen instead of one long scroll. Brand and Colour still show both rooms side by side; component states sit on one labelled canvas with its own light and dark switch.
- Icon buttons show a `Tooltip` instead of a native `title`, and ghost buttons hover to `hover`, because `plate-2` equals `rail` in the light room and the fill disappeared there.
- The segmented control's chosen plate slides between options over 200 ms. It jumps on first render, resize, keyboard changes, and reduced motion.

## Verification

- `pnpm verify` passed.
- `pnpm test:e2e:chromium` passed (17 tests). WebKit and mobile were not run; the tooltip touches focus and touch behaviour, so `pnpm test:e2e` is the complete check.
- Opened the design pages in the local browser at desktop and phone widths, in both rooms.


apps/web/src/client/design/IconToggle.tsx
```diff
@@ -0,0 +1,83 @@
...
+        ref={(node) => {
+          tip.triggerRef.current = node;
+        }}
+        type="button"
+        aria-label={option.label}
+        aria-pressed={on}
+        {...tip.handlers}
+        onClick={(e) => onSelect(e.detail === 0)}
+        className={clsx(
+          "relative grid size-6 place-items-center rounded-full transition-colors duration-150",
```
krambertech, line 74: The new theme choices are only 24×24 px (size-6) in the phone header and component canvases; I measured the rendered buttons in the iPhone/WebKit viewport. This misses Lymi’s 44 px touch-target floor, so please give each option a non-overlapping 44×44 px hit area.

apps/web/src/client/design/DesignLayout.tsx
```diff
@@ -0,0 +1,130 @@
...
+          </IconButton>
+          <Home className="me-auto" />
+          <IconToggle label="Theme" value={theme} onChange={setTheme} options={ROOM_THEMES} />
+        </header>
+        <main className="@container mx-auto w-full max-w-[1120px] flex-1 px-5 pt-8 pb-16 @3xl/shell:px-10">
+          <Outlet />
+        </main>
+      </div>
+
+      <Sheet open={menu} onOpenChange={setMenu} title="Design system">
```
krambertech, line 58: The new mobile Contents drawer is exposed as an unnamed dialog in WebKit: aria-labelledby points to no element, while the visible Design system heading has no id. SheetPanel passes id={undefined} into Drawer.Title, overriding the generated id; please omit that prop when it is undefined or wire the heading id explicitly.

Issue #70, opened 2026-09-13
Finish shared-deck member management


krambertech:
## Outcome

An owner can manage who studies a shared deck. Members can tell that its content is read-only, leave safely, and later resume their own preserved progress when rejoining is allowed.

## Remaining scope

- Show the owner an owner-only member list with joined dates and Remove controls, without exposing review data.
- Show a shared deck's owner in Library and remove owner-only editing actions from member surfaces.
- Keep member content writes forbidden through the app, API, and MCP while allowing members to review only their own states.
- Let a member leave and rejoin through an active link with progress intact; keep an owner-removed account blocked from the generic link.
- Record leave and removal in Activity without exposing private learning state.

## Done when

- Only the owner can list or remove members.
- A member can read the deck, play audio, and grade their own reviews, but cannot change shared content.
- Leaving removes access without deleting states or reviews, and allowed rejoining restores them once.
- Owner removal blocks generic-link re-entry.
- The complete two-learner journey and authorization boundaries pass focused browser and service checks.

## Product context

- [Shared-decks proposal](https://github.com/krambertech/lymi/blob/main/docs/proposals/shared-decks.md)
- [ADR 0011](https://github.com/krambertech/lymi/blob/main/docs/adr/0011-a-shared-deck-is-one-deck-with-many-learners.md)

PR #74, opened 2026-09-13
feat(web): share a deck through a join link

krambertech:
## Summary

Slice 1 of [issue 70](https://github.com/krambertech/lymi/issues/70): share and join.

- Deck settings has a Sharing choice: **Private** (**Link off** once anyone has joined, since members stay) or **Shared by link**. Choosing shared makes the link and shows it under that option with Copy and **Turn off link**. Going back to private asks first, inline, because the URL dies for good; sharing again makes a new URL, and members stay.
- `my.lymi.app/join/<token>` shows the owner, deck name, card count, language, and three recent cards. The Worker picks the state (live, off, archived, invalid), writes the title and Open Graph tags without any card content, and returns 200, 410 or 404. ADR 0011 now records showing example cards, and "no cards before joining" as the rejected option.
- A signed-out classmate taps **Join with Google**. The link rides through sign-in in a ten-minute HttpOnly cookie. `user.create.before` lets that account in without `ALLOWED_EMAILS`, and `session.create.after` finishes the join. Signed-in visitors get **Join**, and members get **Open deck**.

## Screenshots

| Private | Shared by link | Turning the link off |
| --- | --- | --- |
| Sharing set to private | Sharing by link, link masked | Turn-off confirmation |

| Join page, phone, light | Join page, phone, dark |
| --- | --- |
| Join page in light theme | Join page in dark theme |

The join link is masked in the settings images.

## Verification

- `pnpm verify`: passed (143 web unit tests, including new `invitations.test.ts` for every link state, the preview's example cards, a join racing a revocation, owner-only access, concurrent turn-on and join, leave, rejoin and removal)
- `pnpm deploy:check`: passed on the first commit
- `pnpm test:e2e:chromium`: 18 passed, including the new `shared-deck-join.spec.ts`
- `shared-deck-join.spec.ts` passed again in Chromium and WebKit (iPhone) after the redesign. The full Chromium suite ran on the first commit; the rest of the WebKit suite was not run.
- Checked by hand in the browser: the Sharing choice on desktop and phone in light and dark, the turn-off confirmation, and the join page's owner, signed-out, visitor-joins and invalid-link states

## Risks

- Google sign-in is not covered locally. The browser test uses the local email form, which runs the same cookie and hooks. The first live Google join is the real check for the callback path.
- Workers Logs record request paths, so join tokens reach Cloudflare's invocation logs. Nothing in app code logs them. Moving the token out of the path would change the URL shape ADR 0011 chose.
- `join` writes membership only while its link is on and the deck is active, so turning the link off mid-join wins. Concurrent repeats make one membership and one audit row, and turning the link off writes its audit row in the same batch.

## Follow-ups

Slice 2 (member list, read-only member surfaces, Leave, and Activity rows for the access changes) is next. The audit rows for turning the link on and off are already written.


apps/web/src/client/views/DeckSettingsView.tsx
```diff
@@ -205,3 +223,132 @@ export function DeckSettingsView({
...
+          <legend className="sr-only">
+            <Trans>Who can study this deck</Trans>
+          </legend>
+          <RadioCard
+            name={name}
+            value="private"
+            checked={!shared}
+            onChange={() => setConfirming(true)}
+            title={t`Private`}
+            description={t`Only you study this deck.`}
```
krambertech: [P1] Do not call link-off private

Turning this option on does not make the deck private: everyone who already joined keeps access, yet the selected state says ‘Only you study this deck’ and the group is named ‘Who can study this deck.’ That can make an owner believe access was revoked when it was not; model this as whether new people may join, or change the behavior if ‘Private’ is meant literally.

apps/web/src/server/services/invitations.ts
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Join the deck a working link points to. Repeats are safe; a learner the owner removed is
+ * refused, and a link that is off or on an archived deck admits nobody.
+ */
+export async function joinThroughLink(ctx: ServiceContext, token: string) {
+  const link = await linkByToken(ctx.db, token);
+  if (!link || link.revokedAt || link.deckArchivedAt) {
+    throw new ServiceError("not_found", "This join link does not work");
+  }
+  const { role } = await join(ctx, link.deckId, { invitationId: link.id });
```
krambertech, line 215: [P1] Make link validity part of the membership write

The link is checked before `join()`, but the membership write does not depend on the invitation still being active. If the owner turns the link off after this read and before `join()`’s batch, the revocation can complete and this request can still add a new member afterward; condition the membership transition atomically on `revoked_at is null` and the deck still being active.

apps/web/src/server/services/invitations.ts
```diff
@@ -0,0 +1,211 @@
...
+    .set({ revokedAt: now, updatedAt: now })
+    .where(
+      and(
+        eq(schema.deckInvitations.deckId, deckId),
+        eq(schema.deckInvitations.kind, "link"),
+        isNull(schema.deckInvitations.revokedAt),
+      ),
+    )
+    .returning({ id: schema.deckInvitations.id });
+  if (revoked.length > 0) {
```
krambertech: [P2] Keep revocation and its audit recoverable

The permanent revocation commits before the audit write. If `audit()` fails, retrying cannot repair the missing record because no active row remains, so an access change can disappear from Activity; write the revocation and its conditional audit in one atomic or otherwise recoverable operation, as `turnOnJoinLink()` already does.

krambertech (commented):
I found three remaining issues at the current head: the settings state overstates what link revocation does, a concurrent join can cross a completed revocation, and the revocation audit is not recoverable. Focused invitation tests and the Chromium/WebKit join journey pass, but they cover the sequential paths rather than these failure and interleaving cases.

PR #131, opened 2026-09-13
feat(ui): establish the shadcn Base UI foundation

### LINKED ISSUE #119: Establish the shadcn Base UI foundation
## Outcome

Lymi can add generated shadcn components backed by Base UI without adopting the default shadcn visual theme.

## Scope

Add `components.json`, `@base-ui/react`, the current shadcn composition utility, aliases, and the `components/ui` boundary. Configure React without RSC, TypeScript, the existing Tailwind entry point, and Lucide. Generate one low-risk component as a build proof and document the local ownership and update approach in `DESIGN.md`.

## Acceptance

Existing screens render unchanged, Lymi tokens remain authoritative, generated imports resolve in the app, and the pull request records the generated source version or date.

## Verification

- Update the canonical states on the local `/design` route.
- Run the closest affected Playwright journey first.
- Run `pnpm check`, `pnpm typecheck`, `pnpm test`, and `pnpm build`.
- Inspect affected mobile and desktop states in light and dark themes.

## Context

Accepted direction, delivery plan, and full issue set: #118.

This issue does not authorize deployment.

krambertech:
## Summary

- Sets up shadcn on Base UI in `apps/web` (`components.json` with the `base-nova` style from shadcn CLI 4.21.0, `@base-ui/react` 1.8.0, `cn`, and a `components/ui` boundary), with a generated Separator as the build proof. Lymi tokens stay authoritative: none of shadcn's theme variables are added, only the `data-open:`/`data-horizontal:` state variants its generated classes need.
- Adds [ADR 0017](docs/adr/0017-interface-primitives-are-shadcn-components-on-base-ui.md): compound shadcn APIs at call sites, and one device rule for overlays, so menus, forms and confirmations become drawers on touch. The proposal and plan from #118 now link it and name the `ResponsiveMenu` and `ResponsiveDialog` composites.
- Adds component tests in Vitest browser mode (desktop Chromium, touch Chromium, touch WebKit) to `pnpm test`, and a CI step that installs those browsers.

Closes #119.

## Verification

- `pnpm verify` passed, including the Separator test and a state-variant test in all three browsers. The variant test fails on the copied `data-selected` selector and passes with the fix.
- No rendered screen changes; nothing imports the new components yet.

## Risks

- The component tests add a Chromium and WebKit install to the base CI job.


krambertech (commented):
One foundation issue to fix before merge: the selected-state variant does not match Base UI's actual presence attribute. The rest of the foundation looks coherent, and the component suite passed across all three browser instances. Please also remove the generated-with-Claude footer from this PR description per the repository's PR guidance.

apps/web/src/client/styles.css
```diff
@@ -371,6 +371,62 @@
...
+  }
+}
+@custom-variant data-unchecked {
+  &:where([data-state="unchecked"]),
+  &:where([data-unchecked]:not([data-unchecked="false"])) {
+    @slot;
+  }
+}
+@custom-variant data-selected {
+  &:where([data-selected="true"]) {
```
krambertech: Base UI emits `data-selected` as an empty presence attribute, so this selector never matches selected Select or Combobox items and their state styles fail silently. Match `[data-selected]:not([data-selected="false"])` here, consistent with the other boolean variants.

krambertech: Fixed in 0424755: `data-selected` now matches `[data-selected]:not([data-selected="false"])`. `state-variants.browser.test.tsx` checks every copied boolean variant against an empty presence attribute and against `="false"`; it failed on the old selector in all three browsers.

Issue #82, opened 2026-09-13
Publish legal and support pages for public integrations


krambertech:
## Goal

Publish the public trust and support pages needed before Lymi can accept learners through public integrations or directory listings.

The pages must describe the product that actually runs. They should make Lymi's handling of account identity, lesson material, cards, review history, AI processing, audio, API access, and MCP access understandable without promising controls or guarantees the product does not provide.

## Delivery

- [ ] Choose the publisher name and public support contact that should appear consistently across Lymi and provider listings.
- [ ] Publish a privacy policy covering the data Lymi collects and receives, why it is used, which service providers receive it, retention, and the controls available to a learner.
- [ ] Publish terms covering eligibility, learner content, acceptable use, service availability, intellectual property, suspension, termination, and the governing publisher identity.
- [ ] Publish a support page with a working contact route and enough information to troubleshoot sign-in, connected apps, API access, and MCP access.
- [ ] Document a working route for account-data access, export, correction, and deletion while any of those controls remain non-self-service.
- [ ] Audit the implementation and operational configuration behind every statement, including Google sign-in, Cloudflare storage and logs, AI and speech providers, connected-app tokens, push notifications, and beta-signup data.
- [ ] Link the pages from the public site, sign-in and consent surfaces, and relevant API and MCP documentation.
- [ ] Advertise the policy and terms URLs through OAuth protected-resource metadata where supported.
- [ ] Obtain appropriate legal review before treating the text as ready for public users.

## Done when

- Privacy, terms, and support have stable public HTTPS URLs with no authentication requirement or placeholder content.
- The pages use one publisher identity and one working support contact.
- Every described data category, recipient, retention rule, and learner control has been checked against the current product.
- A learner can find the pages before signing in and while managing a connected app.
- The URLs are ready to provide to OpenAI and Anthropic during directory submission.
- Relevant route and link checks pass, followed by `pnpm check`, `pnpm typecheck`, `pnpm test`, and `pnpm build`.

## Not in scope

Opening public registration, submitting or publishing a directory listing, building an interactive MCP App, choosing a paid plan, or replacing legal review with generated text.

PR #135, opened 2026-09-13
feat(mcp): prepare the MCP server for directory review

krambertech:
## Summary

- Disconnecting an app in Settings now blocks its next MCP request, even when its access token is still valid. Each request checks the learner's consent, and the narrower of the token's and the consent's scope applies.
- Tools declare `securitySchemes` at the top level and in `_meta`, plus every hint directory review asks for. Edits are destructive, and only `add_cards` and `update_settings` are idempotent, since other writes add to Activity on every call. A write on a read-only grant returns a step-up challenge. Read tools no longer create a settings row. Unexpected errors reach neither the assistant nor the log in full, and card output drops `createdBy` and `updatedAt`.
- Adds `/.well-known/openai-apps-challenge` for OpenAI domain verification, corrects the ChatGPT setup docs, and adds `docs/mcp-directory-review.md`. Submission still waits on a public account path and #82.

## Screenshots

ChatGPT setup requirements

ChatGPT troubleshooting

MCP overview disconnect note

## Verification

- `pnpm verify` passed.
- `pnpm deploy:check` passed on the first revision.
- Local OAuth with a real consent: writes worked on a write grant, and the same token got a 401 challenge right after Disconnect.

## Risks

- Every `/mcp` request adds one D1 read.
- The SDK has no hook for a top-level `securitySchemes`, so `withSecuritySchemes` wraps its private `tools/list` handler. It throws if the handler moves, and `index.test.ts` checks the field over HTTP.
- `getSettings` now returns in-code defaults for a learner with no row. `settings.test.ts` compares them with a stored row, so a new column default must be added in both places.


apps/web/src/server/mcp/server.ts
```diff
@@ -393,25 +396,44 @@ export function buildMcpServer(principal: McpPrincipal): McpServer {
...
 
   return server;
 }
 
-const read = { readOnlyHint: true, destructiveHint: false, openWorldHint: false };
-const write = { readOnlyHint: false, destructiveHint: false, openWorldHint: false };
+/** The SDK drops a top-level `securitySchemes`; ChatGPT also reads it from `_meta`. */
+const readTool = {
+  annotations: { readOnlyHint: true, destructiveHint: false, openWorldHint: false },
+  _meta: { securitySchemes: [{ type: "oauth2", scopes: ["read"] }] },
```
krambertech, line 435: The wire descriptor contains only `_meta.securitySchemes`, which OpenAI documents as a backward-compatibility mirror of the top-level `securitySchemes`. Serialize both fields; otherwise the current directory scanner may treat the tools as having no declared authentication policy.

apps/web/src/server/mcp/server.ts
```diff
@@ -393,25 +396,44 @@ export function buildMcpServer(principal: McpPrincipal): McpServer {
...
+  overwrites = false,
+}: {
+  idempotent: boolean;
+  overwrites?: boolean;
+}) {
+  return {
+    annotations: {
+      readOnlyHint: false,
+      destructiveHint: overwrites,
+      idempotentHint: idempotent,
```
krambertech, line 453: Card/deck updates and archive/restore operations append audit records and advance `updatedAt` every time. Repeating identical arguments therefore has additional user-visible effects in Activity; mark these tools non-idempotent or make unchanged/repeated writes true no-ops.

apps/site/src/components/docs/pages/McpChatgpt.tsx
```diff
@@ -19,9 +19,13 @@ export function McpChatgpt() {
 
       <H2>ChatGPT</H2>
       <p>
-        Custom connectors need developer mode, and they need a plan that allows them: Pro, Plus,
-        Business, Enterprise or Education. On a managed workspace an admin has to permit custom
-        connectors first.
+        Custom connectors need developer mode, which OpenAI offers on the web for Plus, Pro,
```
krambertech: Current OpenAI guidance limits full MCP write support to Business and Enterprise/Edu, while Pro custom apps are read/fetch-only. It also says ChatGPT may request confirmation depending on permissions and context, not that every write is always confirmed; split the plan capabilities and soften the guarantee.

apps/web/src/server/mcp/server.ts
```diff
@@ -393,25 +396,44 @@ export function buildMcpServer(principal: McpPrincipal): McpServer {
...
   );
 
   return server;
 }
 
-const read = { readOnlyHint: true, destructiveHint: false, openWorldHint: false };
-const write = { readOnlyHint: false, destructiveHint: false, openWorldHint: false };
+/** The SDK drops a top-level `securitySchemes`; ChatGPT also reads it from `_meta`. */
+const readTool = {
+  annotations: { readOnlyHint: true, destructiveHint: false, openWorldHint: false },
```
krambertech, line 434: `list_decks` and `get_settings` call `getSettings()`, which inserts a settings row when none exists. A read-scoped connection can therefore mutate state despite this hint and the product rule that writes require write access; make the getter side-effect-free or initialize settings elsewhere.

apps/web/src/server/mcp/server.ts
```diff
@@ -441,29 +463,50 @@ export function withAiSourceDefaults<T extends SourcedFields>(input: T): T {
...
+  principal: McpPrincipal,
+): Promise<CallToolResult> {
   try {
     return await fn();
   } catch (err) {
-    if (err instanceof ServiceError) return failure(err.message, err.details);
-    throw err;
+    if (err instanceof ReadOnlyConnection) return readOnlyFailure(principal);
+    if (err instanceof ServiceError) return failure(err.message);
+    console.error(err);
```
krambertech: `console.error(err)` can expose information the response correctly hides. Drizzle errors include SQL and bound parameters, potentially leaking card text, notes, and account identifiers into Worker logs; emit only a sanitized error category and non-sensitive correlation ID.

Issue #75, opened 2026-09-13
Add card images and cue-target review modes


krambertech:
## Outcome

Add one private picture per card and explicit cue-target review modes without creating a separate picture-card model.

## Status

Completed through:

- #146: cue-target review modes
- #147: private picture ingestion, delivery, and API and MCP support
- #148: picture review and card rendering
- #204: add and edit every card field, including pictures

## Product context

- [Card-images proposal](https://github.com/krambertech/lymi/blob/main/docs/proposals/card-images-and-visual-review.md)
- [ADR 0014](https://github.com/krambertech/lymi/blob/main/docs/adr/0014-review-modes-use-explicit-cues-and-targets.md)

krambertech:
Implemented across #146, #147, #148, and #204: cue-target review modes, private picture ingestion and delivery, picture review and card rendering, and picture editing are now on main. The proposal and ADR 0014 retain the durable product and architecture rationale.

PR #204, opened 2026-09-14
feat(web): add and edit every card field, pictures included

krambertech:
## Summary

- Every field the API takes on a card can now be set in the app: picture and description, example, pronunciation, notes, language, source, tags and how it's asked. Adding and editing share one form.
- Desktop shows term, meaning and deck, with the rest under More fields. A phone shows a row of field chips, each opening its own drawer. Create more (off by default) keeps the sheet open, and closing a filled sheet offers Undo.
- The card page no longer edits in place; the pencil opens the same form. Part of #75.

## Screenshots

| Desktop | Desktop, more fields |
| --- | --- |
| Add a card on desktop | More fields open |

| Phone | Phone, one field |
| --- | --- |
| Add a card on a phone | Source chip opened |

## Verification

- `pnpm verify`: passes.
- Full browser suite in Chromium: 38 of 38 pass. `deck-creation`, `word-detail` and `core-learning-flow` also pass 22 of 22 in WebKit.
- `deck-page` from #214 tested in-place meaning editing; it now opens the card form instead.

## Risks

- A picture uploads after its card exists. If the upload fails the card stays added without it; the message offers Add picture, which opens the card's edit sheet at the picture field.


apps/web/src/client/components/edit-card-sheet.tsx
```diff
@@ -0,0 +1,65 @@
...
+      // Whatever landed before the failure stays; saving again sends only what is still different.
+      await refreshAfterCardWrite(qc, card.id);
+      return { status: "failed", message: errorMessage(e) };
+    } finally {
+      setPending(false);
+    }
+  };
+
+  return (
+    <Dialog open={!!card} onOpenChange={(open) => !open && onClose()}>
```
krambertech: **Major.** Closing the edit sheet with unsaved changes discards them silently. Verified at 1280×800: change the meaning, press Escape, the sheet closes and no toast appears; on a phone a swipe or a tap on the scrim does the same. The add sheet gets "Card discarded · Undo", so the same slip on an edit is the one place the form breaks "cheap mistakes". Fix: pass `onDraftChange` here too, compare the draft with `card` on close, and when they differ show "Changes discarded" with Undo that reopens the form with `draft`, as `AddCardSheet` already does.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/field-chip.tsx
```diff
@@ -0,0 +1,148 @@
...
+              aria-invalid={invalid || undefined}
+              className="shrink-0 text-muted aria-invalid:text-danger aria-expanded:bg-hover aria-expanded:text-text"
+            >
+              <Icon />
+            </IconButton>
+          )
+        }
+      />
+      <DrawerVirtualKeyboardProvider>
+        <DrawerContent>
```
krambertech: **Major (touch).** The chip's drawer opens with focus on the drawer popup, not on the field: on the iPhone 15 profile, `document.activeElement` after tapping Example is a `DIV`. So every optional field costs two taps and the keyboard does not rise with the drawer, on the machine this layout exists for. Fix: give `DrawerContent` an `initialFocus` that resolves to the first `input, textarea` inside it (the same hook `DialogContent` already exposes).

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/field-chip.tsx
```diff
@@ -0,0 +1,148 @@
...
+            <FilledChip
+              icon={Icon}
+              thumbnail={thumbnail}
+              aria-label={`${label}: ${valueText ?? (typeof value === "string" ? value : label)}`}
+              aria-invalid={invalid || undefined}
+            >
+              {value}
+            </FilledChip>
+          ) : (
+            <IconButton
```
krambertech: **Major (discoverability).** An empty chip is only its icon, so a fresh add sheet on a phone shows seven unlabeled glyphs: a text-quote for Example, a bookmark for Source, a notebook for Notes, sliders for Card settings. Tooltips never show on touch, so the only way to learn what a chip is is to open it. Nothing else in Lymi asks the learner to decode an icon row; DESIGN's "standard controls" rule and the Anki anti-reference both point the other way. Fix: render the empty chip as the same `plate-2` chip with icon and label ("Example", "Notes"), and let the row scroll as it already does once filled; the mock in the proposal shows the row fits three or four labeled chips before the fade.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/card-form.tsx
```diff
@@ -0,0 +1,788 @@
...
+                transition={
+                  reduceMotion
+                    ? { duration: 0.15, height: { duration: 0 } }
+                    : { duration: 0.26, ease: [0.22, 1, 0.36, 1], opacity: { duration: 0.18 } }
+                }
+                // Room on every side, so the clip while it opens never cuts a focus ring.
+                className="-m-1 -mt-3 overflow-hidden p-1"
+              >
+                {/* Only the extra fields scroll on a short screen, so the term and the actions stay put. */}
+                <div className="-mx-1 grid max-h-[max(16rem,calc(92dvh-29rem))] gap-4 overflow-y-auto px-1 pt-2 pb-1 [scrollbar-color:var(--edge-2)_transparent] [scrollbar-width:thin]">
```
krambertech: **Major.** The nested scroll region fails on ordinary laptop heights. At 1280×800 the region is ~270 px, the Pronunciation/Source row is cut at its edge and no scrollbar shows, so Tags, Notes, Language and How it's asked are invisible until the learner discovers the inner scroll. At 1280×620 the `16rem` floor is taller than the room left, so `DialogContent`'s own `overflow-y-auto` takes over and the Save row is off-screen: two nested scrolls, primary action hidden. Fix: drop the inner `max-h`/`overflow` and let the dialog scroll as one column, with the action row `sticky bottom-0 bg-plate` so Term stays at the top and Add/Save stays in reach.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/card-form.tsx
```diff
@@ -0,0 +1,788 @@
...
+    draftRef.current?.({ ...values, languageChosen: chosenLanguage !== null });
+  });
+
+  const clear = (key: string) => setInvalid(({ [key]: _, ...rest }) => rest);
+  const panelProps = (name: Panel) => ({
+    open: panel === name,
+    onOpenChange: (open: boolean) => setPanel(open ? name : null),
+  });
+  // Enter in a one-line field inside a panel is done with that field, not the card.
+  const closeOnEnter = (e: React.KeyboardEvent) => {
```
krambertech, line 270: **Minor.** On the desktop layout Enter in Source or Pronunciation is swallowed: there is no panel to close, so nothing happens, while Enter in Term or Meaning adds the card (verified). Apply `closeOnEnter` only when `chips` is true so the keyboard behaves the same across the whole form.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/card-form.tsx
```diff
@@ -0,0 +1,788 @@
...
+  onCancel,
+  onCreateDeck,
+  onSubmit,
+  onDraftChange,
+  layout,
+  static: st,
+}: CardFormProps) {
+  const { t, i18n } = useLingui();
+  const desktop = useDesktop();
+  const chips = (layout ?? (desktop ? "whole" : "chips")) === "chips";
```
krambertech, line 145: **Minor.** `chips` follows the live media query while `Dialog` freezes its shape for the life of the sheet, so a window dragged across 768 px mid-edit shows the whole layout, Cancel included, inside a drawer, or the chip row inside a centred dialog. Read the shape from the Dialog's context (export a `useDialogShape`) or compute `layout` in the sheets with `useOverlayShape(open)` and pass it down.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/add-card-sheet.tsx
```diff
@@ -21,210 +23,101 @@ interface Props {
...
+      // The card is in either way; a refused picture is said beside it, not instead of it.
+      let pictureError: string | undefined;
+      await savePicture(outcome.card, values).catch((e: unknown) => {
+        pictureError = errorMessage(e);
+      });
+      await refreshAfterCardWrite(qc);
+      const term = outcome.card.term;
+      if (createMore) return { status: "added", term, pictureError };
+      close(false);
+      toast.add({
```
krambertech, line 90: **Minor.** "Added … without its picture" is honest but leaves no way back: the learner has to find the card, open it, press Edit, open Picture, and choose the file again. Give the toast an action, "Add picture", that opens `EditCardSheet` for the returned card with the picture panel open; with Create more on, keep the picture draft in the form instead of clearing it so a retry is one press.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/field-chip.tsx
```diff
@@ -0,0 +1,148 @@
...
+      <DrawerVirtualKeyboardProvider>
+        <DrawerContent>
+          <DrawerHeader className="px-4 pt-1">
+            <DrawerTitle className="text-base">{label}</DrawerTitle>
+          </DrawerHeader>
+          <div className="grid min-w-0 grid-cols-[minmax(0,1fr)] gap-3 px-4 pt-3 pb-5">
+            {children}
+            <div className="flex justify-end">
+              <Button size="sm" onClick={() => onOpenChange(false)}>
+                <Trans>Done</Trans>
```
krambertech: **Minor.** Done is a 56×32 secondary button at the end of the row, while DESIGN has drawer actions stacked full width at 44 px in reach of the thumb. Use the `DialogFooter` treatment (`h-11 w-full`) so the field drawers match every other drawer.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/field-chip.tsx
```diff
@@ -0,0 +1,148 @@
...
+});
+
+/**
+ * The row the chips sit in: one line that scrolls sideways once filled chips outgrow it, fading at
+ * the edge that has more.
+ */
+export function FieldChipRow({ label, children }: { label: string; children: ReactNode }) {
+  const ref = useRef<HTMLDivElement>(null);
+  const [more, setMore] = useState(false);
+  useEffect(() => {
```
krambertech, line 131: **Minor.** This effect has no dependency list, so every keystroke in the form disconnects the `ResizeObserver` and the scroll listener and creates them again. Give it `[]` and observe the row itself; a chip changing width already resizes the row's scrollWidth, which the observer on `el` catches.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/field-chip.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const filled = value !== undefined && value !== null && value !== "" && value !== false;
+  return (
+    <Drawer open={open} onOpenChange={onOpenChange} showSwipeHandle>
+      <DrawerTrigger
+        render={
+          filled ? (
+            <FilledChip
+              icon={Icon}
+              thumbnail={thumbnail}
+              aria-label={`${label}: ${valueText ?? (typeof value === "string" ? value : label)}`}
```
krambertech: **Minor (a11y).** The filled picture chip is announced as "Picture: Picture" because `value` is the label. Pass `valueText` with the file name or the link's host so the name says what was chosen.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/card-picture-field.tsx
```diff
@@ -0,0 +1,296 @@
...
+    value.kind === "file" ? (
+      value.file.name || t`Pasted picture`
+    ) : value.kind === "link" ? (
+      <span className="font-mono text-sm">{value.host}</span>
+    ) : (
+      t`Current picture`
+    );
+  const caption =
+    value.kind === "file"
+      ? t`${formatSize(value.file.size, i18n.locale)} · sent when you save`
```
krambertech: **Minor (copy).** "sent when you save" sits under a button that says "Add to Lezione 12" in the add form. "Sent with the card" reads true in both modes.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/card-form.tsx
```diff
@@ -0,0 +1,788 @@
...
+      <FieldError>{invalid.example}</FieldError>
+    </Field>
+  );
+
+  const pronunciationField = (
+    <Field>
+      <FieldLabel className={chips ? "sr-only" : undefined}>{t`Pronunciation`}</FieldLabel>
+      <Input
+        value={pronunciation}
+        maxLength={200}
```
krambertech: **Minor.** `maxLength` here and on Source truncates silently, and `TagsInput` slices a tag to 40 characters, while every other field validates on submit with a sentence under the control. The form already carries the schema messages for these fields, so drop the hard limits and let the same rule speak everywhere.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/web/src/client/components/add-card-sheet.tsx
```diff
@@ -21,210 +23,101 @@ interface Props {
...
+          onCreateMoreChange={(on) => {
+            setCreateMore(on);
+            rememberCreateMore(on);
+          }}
+          onCancel={() => close(true)}
           onCreateDeck={
             onCreateDeck &&
             (() => {
-              onOpenChange(false);
+              close(false);
```
krambertech: **Minor (pre-existing, now fixable).** "New deck" from an account with no decks calls `close(false)`, so a typed term is dropped without Undo, while the field says "Create the first one and this card goes in it." The draft mechanism this PR adds could keep that promise: hold the draft, and when the deck lands reopen the sheet with it and the new deck selected.

---
_Generated by [Claude Code](https://claude.ai/code)_

krambertech (commented):
## Verdict

**Changes needed.** The data path is sound and the desktop happy path reads well, but four rendered findings regress "cheap mistakes" and reach on the phone, the layout this PR was built for: an edit closed by mistake is lost silently, a chip's drawer opens without focusing its field, the empty chip row is seven unlabeled icons, and the desktop sheet's nested scroll hides the Save row at laptop heights.

## What holds up

- `addInput` / `cardPatch` / `savePicture` are the right split: the patch carries only what changed, so History stays honest, and the picture rides after the card exists. The unit tests cover the edge that matters (`null` review modes follow the deck again).
- Client validation reuses the route's schema and `revealsAnswer` moved into `core`, so the description rule cannot drift between the form and the API.
- Refused fields open where they can be fixed on both machines (verified: the description error unfolds More fields and focuses the input on desktop; on the phone it opens the Picture drawer with the message).
- The "Added … without its picture" path is truthful in the notice and the toast, and the card still lands.
- The card page as a read-only page with the pencil in the bar is a clear improvement over turning text into fields.

## Findings

Inline, one per thread. Majors: unsaved edits discarded without Undo (`edit-card-sheet.tsx`), field drawer does not focus its field on touch (`field-chip.tsx`), empty chips are icon-only (`field-chip.tsx`), nested scroll hides the action row on short windows (`card-form.tsx`). Minors: Enter swallowed in Source/Pronunciation on desktop, live layout switch inside a frozen dialog shape, no way back from a failed picture, Done button size in field drawers, effect without deps, "Picture: Picture" name, "sent when you save" copy, silent `maxLength`, dropped draft on "New deck".

One suggestion outside the findings: with Example, Pronunciation and Source being the fields a lesson actually produces, Picture as the first and largest thing under More fields (and the first chip) pushes them down. Example first, Picture after Notes, then the hairline and settings, would match how often each is used.

## Coverage

Diff of all 28 files read; rendered add, edit, discard/Undo, duplicate, refused-description and Create more flows on the PR head against a local D1 with the `learner` persona. Desktop Chrome 1280×800 and 1280×620, light and dark; iPhone 15 profile (393×852, touch) light and dark; iPad Mini 1024×768 touch, one shot. Not inspected: WebKit rendering, a real software keyboard, drag-and-drop and paste of a picture, a successful upload's thumbnail (the test file was refused by `inspectImage`, which exercised the failure path instead), RTL.

## Verification

- CI on `ce51840`: plan, quality, browser E2E (Chromium), app and site previews all green.
- Local: `pnpm install`, `pnpm db:migrate`, Vite dev server, two Playwright scripts (Chromium 1194) driving the flows above; no console or page errors beyond the expected 400 from the refused test image.
- Not run locally: `pnpm verify` (CI's quality job is the record), WebKit E2E.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #248, opened 2026-09-15
feat(content): add the Everyday Estonian deck and loader

krambertech:
## Summary

- `content/decks/everyday-estonian.json` is the first public deck: 99 cards in eight sections, English meanings, and Markdown notes that explain how phrases are built. Meanings and notes are written for Lymi; the Institute of the Estonian Language A1 list is the cited source.
- `pnpm load-deck <file> --url <origin> [--publish]` creates the deck, its sections and cards through the API as the account `LYMI_API_KEY` belongs to, and refuses to run if that account already has a deck with the same name.

Stacked on #247, which adds publishing.

## Verification

- `node --test scripts/load-deck.test.mjs`: 4 passed, including a check that the Everyday Estonian file has no missing names or repeated terms.
- Loaded the file into a local account with the script: all eight sections created. Some cards were skipped only because that persona already held the same terms.
- `pnpm check`: passed.

## Risks

- The Estonian needs a native read before publishing. Least certain: *Kas kotti soovite?*, *Järgmine peatus …*, and *Ma sooviksin aega perearsti juurde.*


krambertech:
Closing: the deck lives in the database once loaded, so the content file and loader don't belong in the repo. The deck will be loaded directly with the publisher account after #247 ships.

